PR #8, opened 2017-03-01
Add lcov file reading and storage

ryanluker:
## Work:
- [x] cleanup tests
- [x] add new failing tests for gutters class
- [x] add tests for new lcov reading
- [x] add public display coverage function
- [x] add public dispose function
- [x] add private read lcov function


ryanluker:
completes #4

PR #122, opened 2018-03-27
Add simple cache.

spodym:
Hi @ryanluker !

Every switch between tabs is running extraction. I am working with quite big coverage file and this results in considerable lag in displaying coverage. After a while this becomes mildly distracting.

I've seen that you are planning to add some caching in future but please consider this PR as intermediate solution.

### DIFF
--- a/src/gutters.ts
+++ b/src/gutters.ts
@@ -106,7 +106,8 @@ export class Gutters {
             this.loadAndRenderCoverage(textEditor, pickedFile).catch(() => {});
 
             this.coverageWatcher = vscodeImpl.watchFile(pickedFile);
-            this.coverageWatcher.onDidChange((event) => this.renderCoverageOnVisible(pickedFile));
+            this.coverageWatcher.onDidChange(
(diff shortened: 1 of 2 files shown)


ryanluker:
@spodym thanks for the pull request. I dont mind pulling this in and making a dev release of the extension (generating an .vsix file that you would manually install https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix) but the amount of overhead in terms of scheduling testing and then deploying to the marketplace would be too high when I am close to finishing the new architecture. How does the dev release sound to you?

spodym:
I agree that it might not be worth doing marketplace release. Besides in meantime I have managed to install my branch locally so I am good.
I am not sure if anyone else would benefit from this dev release so I leave decision up to you.

BTW. If you need a hand with testing or developing new version - I would be happy to help :)

ryanluker:
@spodym good to hear you got this working locally. Thanks for the offer of help :smile: once I get the tasks broken down and fleshed out a bit more I will find one for us to collaborate on!

PR #138, opened 2018-06-28
Fix crash when reporter throws an error.  Also adds a pretty command to use tslint fix.

brandongk60:
@ryanluker


ryanluker:
@brandongk60 looks good thanks for adding tests too!

PR #139, opened 2018-06-29
#137 - Adding strict null checks compiler option and fixing resulting errors.

### LINKED ISSUE #137: Cleanup function parameters and possible undefined
Based on error reporting version 2.0.0 has had a few crashes in these two functions, mostly undefined errors. 

Check intersection being undefined before using reverse
https://github.com/ryanluker/vscode-coverage-gutters/blob/5eee5f7afd1e6b2b5e0805c96691f97981330246/src/renderer.ts#L125-L142

Check filename / name being undefined before manipulating
https://github.com/ryanluker/vscode-coverage-gutters/blob/5eee5f7afd1e6b2b5e0805c96691f97981330246/src/renderer.ts#L144-L152

## Work:
- [x] failing unit tests for above errors
- [x] add proper ts types for the api (currently assume ```fileName: string``` when maybe it should be ```fileName: string | undefined```)
- [x] handle undefineds

brandongk60:
@ryanluker - I don't think this fixes #137 at this point, but it's a start.

## Issue:
#137 

Turning on typescript `strictNullChecks` option to deal with the undefined issues.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

I had to make a bunch of changes to get this compiling after turning this option on.


src/coverage.ts
```diff
@@ -57,20 +57,26 @@ export class Coverage {
...
-                return [].concat(...coverageInWorkspaceFolders);
+                return new Array().concat(...coverageInWorkspaceFolders);
             });
     }
 
     public load(path: string): Promise<string> {
         return new Promise<string>((resolve, reject) => {
             this.fs.readFile(path, (err, data) => {
-                if (err) { return reject(err); }
+                if (err && Object.keys(err).length !== 0) { return reject(err); }
```
brandongk60: This is a funny way for checking errors, but I can't put undefined / null / etc. in the test with strictTypeChecking turned on.  Maybe it's better to leave this as is and then make a compiler exception in the test?

ryanluker: I tinkered with this a bit and found ```as any``` to be an ok middle ground, then you can go back to the more community standard ```if(err) reject(err)``` as well. https://github.com/ryanluker/vscode-coverage-gutters/commit/1c47fbad3e0b90d0118f46592b70f3ac8cd03971

brandongk60: Thanks @ryanluker.  Merged that patch into this PR.

test/extension.test.ts
```diff
@@ -5,8 +5,20 @@ import {ICoverageLines} from "../src/renderer";
 suite("Extension Tests", function() {
     this.timeout(25000);
 
+    test("Should start extension @integration", async () => {
```
ryanluker: I actually removed this one as it was sometimes randomly failing due to startup times (annoying).

ryanluker:
@brandongk60 just the one comment about the test we can probably remove then this is good to merge.

PR #142, opened 2018-07-02
#137 - Splitting TopSectionFinder from Renderer and adding unit tests to both. (WIP)

brandongk60:
## Issue:
#137 

Splitting up the renderer.ts file to facilitate testing.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,0 +1,34 @@
+
+/**
+ * Finds the matching suffixes of the string, stripping off the non-matching starting characters.
+ * @param base
+ * @param comparee
+ */
+export function findIntersect(base: string, comparee: string): string {
+    const a = [...base].reverse();
+    const b = [...comparee].reverse();
+
+    // find the intersection and reverse it back into a string
+    const intersection: string[] = [];
+    let pos = 0;
+    // stop when strings at pos are no longer are equal
+    while (a[pos] === b[pos]) {
+        // if we reached the end or there isnt a value for that pos
+        // exit the while loop
+        if (!a[pos] || !b[pos]) { break; }
+        intersection.push(a[pos]);
+        pos++;
+    }
+    const subInt = intersection.reverse().join("");
+    return subInt;
+}
+
+export function normalizeFileName(fileName: string): string {
+    let name = fileName;
+    // make file path relative and OS independent
+    name = name.toLocaleLowerCase();
+    // remove all file slashes
+    name = name.replace(/\//g, "###");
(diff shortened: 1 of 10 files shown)


src/renderer.ts
```diff
@@ -20,25 +15,22 @@ export interface ICoverageLines {
...
+        this.topSectionFinder = topSectionFinder;
     }
 
     /**
      * Renders coverage to editors
      * @param sections cached set of sections
      * @param textEditors currently visible text editors
      */
-    public async renderCoverage(
+    public renderCoverage(
```
brandongk60: @ryanluker - I don't think this function needs to be `async`...

src/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function normalizeFileName(fileName: string): string {
+    let name = fileName;
+    // make file path relative and OS independent
+    name = name.toLocaleLowerCase();
+    // remove all file slashes
+    name = name.replace(/\//g, "###");
+    name = name.replace(/\\/g, "###");
+    return name;
+}
```
ryanluker: good idea, helper library makes sense at this point!

src/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    const subInt = intersection.reverse().join("");
+    return subInt;
+}
+
+export function normalizeFileName(fileName: string): string {
+    let name = fileName;
+    // make file path relative and OS independent
+    name = name.toLocaleLowerCase();
+    // remove all file slashes
```
ryanluker: wasnt sure of a node api that would properly strip unix and posix slashes to allow for comparisons... the use case here is windows 10 allows for WSL to generate coverage and have a the extension consume it in windows.

src/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+    const intersection: string[] = [];
+    let pos = 0;
+    // stop when strings at pos are no longer are equal
+    while (a[pos] === b[pos]) {
+        // if we reached the end or there isnt a value for that pos
+        // exit the while loop
+        if (!a[pos] || !b[pos]) { break; }
+        intersection.push(a[pos]);
+        pos++;
+    }
```
ryanluker: this while loop is pretty dangerous and has caused a few bugs. originally I was considering recursion here with a exit condition of getting to the end of one array. it is fine the way it is though as I see you have quite a few more tests backing it thanks!

src/renderer.ts
ryanluker: yep doesnt look that way good catch!

PR #145, opened 2018-07-09
Show coverage only to files that match exactly

jjuzna:
Added logic to show coverage only to files that match exactly (new configuration option matchNameExact)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,11 @@
           "default": "cov.xml",
           "description": "name of your xml file"
         },
+        "coverage-gutters.matchNameExact": {
+          "type": "boolean",
+          "default": false,
+          "description": "match lcov source file only if exactness score = 100"
+        },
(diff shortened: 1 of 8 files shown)


ryanluker:
@jjuzna thanks for putting up a PR! This seems like a good addition but I wonder if we could use a number instead so that way people can decide their threshold for `correctness`. In addition to that change, we could also use some new tests to cover that the topSection properly ignores some scores based on the config store.

jjuzna:
I have no problem with numbers. About tests - sorry I am writing this code in notepad++ and do not have running environment, therefore I cannot execute tests. My contribution has started as updates to extension js source (version 2.0.1).

ryanluker:
@jjuzna closing this PR as we are going away from score based logic in favour of relative paths, thanks for putting the effort in though!

PR #146, opened 2018-07-12
Making a task which rebuilds test coverage for Node

brandongk60:
hard-coding the coverage file in the repo is non-indicative as the path is wrong (i.e. has ryanluker hard-coded).

The Python project matters less because the `cov.xml` file has relative paths anyways.


example/node/lcov.info
```diff
@@ -1,107 +0,0 @@
-TN:
-SF:/home/ryanluker/dev/vscode-coverage-gutters/example/node/test.js
```
brandongk60: These absolute paths are wrong on every machine but @ryanluker's, so we should not rely on them.

src/renderer.ts
```diff
@@ -54,10 +54,10 @@ export class Renderer {
             // find best scoring section editor combo (or undefined if too low score)
             const topSection = this.topSectionFinder.findTopSectionForEditor(textEditor, sections);
 
-            if (!topSection) { return ; }
-
-            this.filterCoverage(topSection, coverageLines);
-            this.setDecorationsForEditor(textEditor, coverageLines);
+            if (topSection) {
```
brandongk60: @ryanluker this can cause the wrong score, etc. to carry over because it doesn't flush properly.  noticed with the Python tests with the score filtering thing.

example/python/cov.xml
```diff
@@ -3,7 +3,7 @@
 	<!-- Generated by coverage.py: https://coverage.readthedocs.io -->
 	<!-- Based on https://raw.githubusercontent.com/cobertura/web/master/htdocs/xml/coverage-04.dtd -->
 	<sources>
-		<source>/home/ryanluker/dev/vscode-coverage-gutters/example/python/foobar</source>
+		<source>/home/brandongk/Desktop/vscode-coverage-gutters/example/python/foobar</source>
```
brandongk60: @ryanluker -> We need to make this file auto-calculate as well, it will fail anywhere except my machine.

src/renderer.ts
```diff
@@ -51,13 +49,14 @@ export class Renderer {
...
-            // find best scoring section editor combo (or undefined if too low score)
-            const topSection = this.topSectionFinder.findTopSectionForEditor(textEditor, sections);
+            if (!textEditor || !textEditor.document) {return; }
 
-            if (!topSection) { return ; }
+            const section = sections.get(normalizeFileName(textEditor.document.fileName));
 
-            this.filterCoverage(topSection, coverageLines);
-            this.setDecorationsForEditor(textEditor, coverageLines);
+            if (section) {
```
brandongk60: @ryanluker this can cause the wrong score, etc. to carry over because it doesn't flush properly. noticed with the Python tests with the score filtering thing.

example/node/lcov.info
ryanluker: these paths are fine in the new way we do path comparison (now that top score is dead). it also has the benefit of the build machine not needed to be able to generate xml for node, python, php, etc.

src/renderer.ts
ryanluker: good catch, fixed this in my other PR #152

ryanluker:
@brandongk60 closing this as we are doing away with the top score logic but replaced it with relative comparison logic.

Issue #160, opened 2018-08-07
Coverage is not visible on VSCode - Flutter


diegoveloper:
I was using this plugin months ago and It was working perfect, now I can't see the coverage of my code. Flutter - Dart.
I'm using the latest version v2.1.1
The coverage is generated under the folder /coverage/lcov.info

diegoveloper:
Log:

[1533851134427][coverageservice]: INITIALIZING
[1533879404563][coverageservice]: LOADING
[1533879408910][coverageservice]: Loading 1 file(s)
[1533879408922][coverageservice]: Caching 66 coverage(s)
[1533879408922][coverageservice]: READY
[1533879408922][coverageservice]: RENDERING
[1533879408937][coverageservice]: READY
[1533879415881][coverageservice]: RENDERING
[1533879415882][coverageservice]: READY
[1533879424919][coverageservice]: LOADING
[1533879427446][coverageservice]: RENDERING
[1533879427447][coverageservice]: READY
[1533879428404][coverageservice]: Loading 1 file(s)
[1533879428409][coverageservice]: Caching 72 coverage(s)
[1533879428409][coverageservice]: READY
[1533879428409][coverageservice]: RENDERING
[1533879428410][coverageservice]: READY
[1533879439563][coverageservice]: LOADING
[1533879442636][coverageservice]: Loading 1 file(s)
[1533879442639][coverageservice]: Caching 72 coverage(s)
[1533879442639][coverageservice]: READY
[1533879442639][coverageservice]: RENDERING
[1533879442640][coverageservice]: READY

looking the log seems like it is working but nothing happens on the IDE

MoxyFruzy:
I have the same issue but with Javascript.

ryanluker:
@diegoveloper thanks for the issue and the extra information, I will hopefully get some free time to take a look at this in the near future. In the meantime (like @MoxyFruzy said) you can manually install one of the previous releases using the vsix isntall method https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix . Also if you are able to confirm that you can see the coverage in the example folder (for this project) that will help narrow down the possibilities. Finally any examples of your coverage files (lcov.info or xml ones) would assist in the investigation too.

diegoveloper:
Ok, looks like version 2.0.0 is working fine for me.  >= 2.0.1 not working at all.
info file attached
[...]

PR #173, opened 2018-10-12
convert relative paths from lcov output to absolute paths

karn09:
When an LCOV file contains paths relative to the project, the opened document will not be highlighted.

For example, transforming the below:
`SF:/home/ryanluker/dev/vscode-coverage-gutters/example/node/test.js`

to:
`SF:./example/node/test.js`

Will cause test.js to not be found, which results in coverage not being highlighted.

This addition attempts to search for absolute paths only when the provided path is relative, otherwise it will return the already known `possiblePath`. 

Instead of searching through the various workspace folder globs, this will inspect the currently open documents made available via `workspace.textDocuments`.

### DIFF
--- a/src/files/coverageparser.ts
+++ b/src/files/coverageparser.ts
@@ -121,12 +121,9 @@ export class CoverageParser {
         }
 
         function findAbsolutePathFromTextDocumentsWhenRelativePath() {
-            
             if (!workspace.textDocuments) { return possiblePath; }
-            if (!possiblePath.startsWith('.')) { return possiblePath }
-            
+            if (!possiblePath.startsWith(".")) { return possiblePath; }
             const partialPath = possiblePath.substring(1);
-        
             // look through currently open documents for possible path
             const filePaths = workspace.textDocuments.map(
                 (document) => document.uri.fsPath,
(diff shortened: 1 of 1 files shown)


karn09:
This may partially address https://github.com/ryanluker/vscode-coverage-gutters/issues/160

Also, the build errors are odd: `Error: Extension `ryanluker.vscode-coverage-gutters` is not known or not activated`

Seems unrelated to these changes.

ryanluker:
@karn09 thanks for the PR! I rebuilt the branch as sometimes the integration / acceptance tests fail due to the container not downloading dependent resources in time, it is passing now.

I really like the way you are detecting the partial paths so I was thinking of modifying this a bit over the weekend to be always used with every parser type and remove my old version. {1}

{1} https://github.com/ryanluker/vscode-coverage-gutters/blob/26742f7d4bd8d84c0cd8c71c416819322ed9cfd9/src/files/coverageparser.ts#L95-L121

ryanluker:
@karn09 going to continue the work on my branch for this, thanks again for kicking off this bug fix :+1: 
https://github.com/ryanluker/vscode-coverage-gutters/pull/177

karn09:
Thanks for taking a look at the PR! I just saw your new update. I had initially thought it was odd that ALL folders were being checked since I was using display gutters on a per file basis, and hadn't considered that this was meant to apply gutters to ALL files. Whether they were open currently or not.

Happy to see that this inspired changes, which totally fixes my original issue!

ryanluker:
@karn09 Yah the display gutters uses the exact some modules as the watch functionality to keep the code dry and reasonable, so when you display coverage on a page it will load the coverage from the file system while the watch feature will reuse the cache (after the first load) for every new file opened.

karn09:
@ryanluker very cool! It took me some to unravel, but that makes a lot of sense. Thanks for explaining!

Issue #175, opened 2018-10-16
Search for jacoco.xml file regardless of configured Xmlname value


samuel-barbosa:
By default, the XML coverage file for Java projects using Jacoco plugin is jacoco.xml.
It'd be good if the Coverage-Gutters plugin could automatically recognize it with no need to change the Xmlname configuration

PR #176, opened 2018-10-16
[#175] Search for jacoco.xml file regardless of configured Xmlname value

samuel-barbosa:


ryanluker:
@samuel-barbosa thanks for the PR! I was thinking of actually combining the lcov and the xml names into one array you get back from the config store (thinking something like `coverageFiles`). This array of files would be defined by default from the settings.json (there you could add your own sets of defaults or use the one I come up with). I will tinker with this of the weekend.

ryanluker:
@samuel-barbosa thanks again for the PR, I have included a more robust solution in https://github.com/ryanluker/vscode-coverage-gutters/pull/181.

Issue #150, opened 2018-07-25
Error Failed to parse string


Licen-it:
I'm always getting this error when i try to use the extension. I tried adding the watch on `lcov.info` and `cobertura-coverage.xml`, but the error is the same and nothing get displayed on the source code.

Here is the screenshot of the extension output:
![image](https://user-images.githubusercontent.com/35496522/43193643-71ba00e8-8ff8-11e8-8dbb-79be669228e9.png)

ryanluker:
@Licen-it thanks for submitting an issue! this is related to the score logic that was added recently and should be fixed in the next version of the extension I am working on (2.1.0). In the meantime you could try just loading lcov.info to see if that will help you til next release.

ryanluker:
@Licen-it hmm the only other thing i would try before 2.1.0 would be to open the example workspace project here in this repo and see if the coverage shows up for you or not.

ryanluker:
@Licen-it 2.1.0 is being released this afternoon (pst) to the marketplace. let me know if this release doesn't fix your issue and we can reopen the ticket and do some more investigating.

ryanluker:
@Licen-it ahh yes the plugin does not look outside the project / workspace folder when it does the collection phase https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/filesloader.ts#L86 this is mainly to prevent looking everywhere in your computer for coverage files but I could imagine a feature where you say where your coverage folder is located and it wouldnt use the glob find in that case. Let me think about this for the 2.1.1 release but in the meantime can you not include your coverage inside your project folder?

Licen-it:
@ryanluker ok...but i was wondering why aren't you using the paths in the coverage files (i can see `lcov.info` for each file provides the full path while `cobertura-coverage.xml` provides the top level folder path and then relative paths to each file)

Licen-it:
Anyway i just checked putting the coverage file on top level folder and still not working...i'm still getting  the `Failed to parse string` error

ryanluker:
[...]

PR #191, opened 2018-11-21
Handle lcov file parsing errors

DW8Reaper:
Running coverage-gutter on my project keeps failing with an error ```Error: Failed to parse string``` similar to what is described in the issue [Error Failed to parse string #150](https://github.com/ryanluker/vscode-coverage-gutters/issues/150). 

For me it turned out to be invalid lcov files in our project but there was no way to remedy this and no log to see which files were in error. I ran both ```npm test-unit``` and ```npm test-integration``` both worked but ```npm test``` fails but looks like that commands setup might not be correct.

### DIFF
--- a/src/files/coverageparser.ts
+++ b/src/files/coverageparser.ts
@@ -31,6 +31,7 @@ export class CoverageParser {
         let coverages = new Map<string, Section>();
 
         for (const file of files) {
+            const fileName = file[0];
             const fileContent = file[1];
 
             // file is an array
@@ -51,7 +52,7 @@ export class CoverageParser {
                     coverage = await this.xmlExtractCobertura(fileContent);
                     break;
                 case CoverageType.LCOV:
-                    coverage = await this.lcovExtract(file[0], fileContent);
+                    coverage = await this.lcovExtract(fileName, fileContent);
                     break;
                 default:
                     break;
@@ -205,7 +206,8 @@ export class CoverageParser {
             try {
                 source(lcovFile, async (err, data) => {
                     if (err) {
-                        this.handleError("lcov-parse", err, filename);
+                        err.message = `filename: ${filename} ${err.message}`;
+                        this.handleError("lcov-parse", err);
                         return resolve(new Map<string, Section>());
                     }
                     const sections = await this.convertSectionsToMap(
@@ -216,18 +218,18 @@ export class CoverageParser {
                     return resolve(sections);
                 });
             } catch (error) {
-                this.handleError("lcov-parse", error, filename);
+                error.message = `filename: ${filename} ${error.message}`;
+                this.handleError("lcov-parse", error);
                 return resolve(new Map<string, Section>());
             }
         });
     }
 
-    private handleError(system: string, error: Error, filename?: string) {
+    private handleError(system: string, error: Error) {
         const message = error.message ? error.message : error;
         const stackTrace = error.stack;
         this.outputChannel.appendLine(
-            `[${Date.now()}][coverageparser][${system}]: Error: ${message}${
-                (!!filename) ? ` in file: "${filename}"` : ""}`,
+            `[${Date.now()}][coverageparser][${system}]: Error: ${message}`,
         );
         if (stackTrace) {
             this.outputChannel.appendLine(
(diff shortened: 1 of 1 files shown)


src/files/coverageparser.ts
```diff
@@ -200,11 +200,14 @@ export class CoverageParser {
         }
     }
 
-    private lcovExtract(lcovFile: string) {
+    private lcovExtract(filename: string, lcovFile: string) {
         return new Promise<Map<string, Section>>((resolve, reject) => {
             try {
                 source(lcovFile, async (err, data) => {
-                    if (err) { return reject(err); }
+                    if (err) {
```
DW8Reaper: Removed the reject since the calling code doesn't have a try catch nor a then for the promise so it fails and no coverage is shown. I think this new change was the intended behavior to return empty coverage for invalid files

src/files/coverageparser.ts
```diff
@@ -213,17 +216,18 @@ export class CoverageParser {
...
             } catch (error) {
-                this.handleError("lcov-parse", error);
+                this.handleError("lcov-parse", error, filename);
                 return resolve(new Map<string, Section>());
             }
         });
     }
 
-    private handleError(system: string, error: Error) {
+    private handleError(system: string, error: Error, filename?: string) {
```
DW8Reaper: Added the file in question to the method signature, really this is just for error reporting so the user can track down the messed up files.

src/files/coverageparser.ts
DW8Reaper: On another note the exact same problem will happen for the other file types as well, I wanted to stick to the lcov for now since I don't really know what I am doing in your project ;)

ryanluker: Yah exposing these errors to the developer via the output channels is a better call, as you could still use the extension with 1 out of 3 coverage files in error for example.

src/files/coverageparser.ts
```diff
@@ -51,7 +51,7 @@ export class CoverageParser {
                     coverage = await this.xmlExtractCobertura(fileContent);
                     break;
                 case CoverageType.LCOV:
-                    coverage = await this.lcovExtract(fileContent);
+                    coverage = await this.lcovExtract(file[0], fileContent);
```
ryanluker: to prep for me applying a similar to fix the other parsers, this (`file[0]`) should be assigned to a new variable near to line 34 where `fileContent` is.

src/files/coverageparser.ts
```diff
@@ -213,17 +216,18 @@ export class CoverageParser {
                     return resolve(sections);
                 });
             } catch (error) {
-                this.handleError("lcov-parse", error);
+                this.handleError("lcov-parse", error, filename);
```
ryanluker: instead of adding filename to the `handleError` signature can we instead construct a new variable that appends the filename to the error message?
```
} catch (error) {
  error.message = `filename: ${filename} ${error.message}`
  this.handleError("lcov-parse", error);
}
```

src/files/coverageparser.ts
```diff
@@ -200,11 +200,14 @@ export class CoverageParser {
...
     }
 
-    private lcovExtract(lcovFile: string) {
+    private lcovExtract(filename: string, lcovFile: string) {
         return new Promise<Map<string, Section>>((resolve, reject) => {
             try {
                 source(lcovFile, async (err, data) => {
-                    if (err) { return reject(err); }
+                    if (err) {
+                        this.handleError("lcov-parse", err, filename);
```
ryanluker: Similar to I mentioned below can we prepend to the error instead, so we dont leak information to the `handleError` function? I am hoping to refactor the handleError function here soon to be even more generic and reusable across the extension so not having specific info like filename will help greatly :smile: .
```
  error.message = `filename: ${filename} ${error.message}`
  this.handleError("lcov-parse", error);
```

src/files/coverageparser.ts
ryanluker: As mentioned above, if we go the prepend to the error message near to the error logic we can revert these changes to the `handleError`.

ryanluker:
@DW8Reaper Thanks again for the pull request! This improvement will go out with the 2.3.0 release but I have also included the latest master vsix build below for you to use in the meantime if you wish.

[vscode-coverage-gutters-2.3.0.zip](https://github.com/ryanluker/vscode-coverage-gutters/files/2614088/vscode-coverage-gutters-2.3.0.zip)

PR #215, opened 2019-04-04
Don't crash if we find empty lcov.info (#150)

ermshiperete:
Sometimes we may encounter an empty `lcov.info` file. Since we know
that an empty file can't contain valid coverage data, we simply
ignore it rather than unsuccessfully trying to load.

---
This change is [Reviewable](https://reviewable.io/reviews/ryanluker/vscode-coverage-gutters/215)

### DIFF
--- a/src/files/coveragefile.ts
+++ b/src/files/coveragefile.ts
@@ -1,4 +1,5 @@
 export enum CoverageType {
+    NONE,
     LCOV,
     CLOVER,
     COBERTURA,
@@ -19,7 +20,7 @@ export class CoverageFile {
      * @param file file to detect type information
      */
     private setFileType(file: string) {
-        let possibleType = CoverageType.LCOV;
+        let possibleType = CoverageType.NONE;
         if (
             file.includes("<?xml") &&
             file.includes("<coverage") &&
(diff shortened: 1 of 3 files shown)


src/files/coveragefile.ts
```diff
@@ -20,7 +21,9 @@ export class CoverageFile {
      */
     private setFileType(file: string) {
         let possibleType = CoverageType.LCOV;
-        if (
+        if (file === "") {
+            possibleType = CoverageType.NONE;
```
ryanluker: can we flip this around to be the default `else` case instead? seems to me that would help with coverage types the system doesnt understand as well! thoughts?

ermshiperete: Done.

ryanluker:
@ermshiperete Thanks for the support and quick fix on this! I think make a dev release (2.4.0) and see if this fixes #150 for the other person as well.

ermshiperete:
How can I run the unit tests?

ryanluker:
@ermshiperete if you mean run the CI tests they auto run with every branch and are visible in azure pipeline https://ryanluker.visualstudio.com/vscode-coverage-gutters/_build/results?buildId=147

if you mean how to you locally run them, you can use  `"test-unit": "cross-env MOCHA_GREP=@unit npm test --silent",` as per the `package.json`

PR #223, opened 2019-06-05
Add config option for context menu commands

### LINKED ISSUE #216: Remove from right-click menu

I don't need quick access to coverage gutter commands. It doesn't need to be in the right click menu; it takes up too much space. The command palette and the status bar are sufficient.

Option to remove context menu entries.

Alternative: Remove context menu entries without option. I assume that some people like it.

**Additional context**

N/A

wenbei999:
There used to be options for this, not sure why it doesn't work anymore. This is the new implementation. 

It used to be several different config flags, but I combined it into one. I figured people would either want all of it or none. It also reduces clutter from too many options. However I noticed you had very fine-grain control for the existing options. Opinions?

Side note, it kind of bothers me that 2 of the config descriptions are Capitalized while the rest are not :(

See links below for details.
https://code.visualstudio.com/api/references/contribution-points#contributes.menus
https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts

Fixes #216.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,11 @@
           "type": "boolean",
           "default": true,
           "description": "enable or disable the status bar item"
+        },
+        "coverage-gutters.customizable.context-menu": {
+          "type": "boolean",
+          "default": true,
+          "description": "enable or disable quick commands in the context menu"
         }
       }
     },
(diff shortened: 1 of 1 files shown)


ryanluker:
@wenbei999 Thanks a bunch for this PR! I am not sure when this feature was lost but good to have it back. I also dont enjoy the mismatch capitalisation and seeming they are just description docs I can fix those up in my other PR (thanks for letting me know about them!).

I also checked it out locally and it works well :shipit:.

ryanluker:
@wenbei999 you can find the description changes here https://github.com/ryanluker/vscode-coverage-gutters/pull/222/commits/410ce5a8e12fc84c4c7163decb5b456d298d802d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R130 thanks again for the PR!

PR #246, opened 2019-11-23
Read multiple tests for the same file in one coverage report

### LINKED ISSUE #245: When multiple sections list the same file (SF), only the last is applied
I have a lcov.info file with the results of many tests run together and this extension seems to only take the last set of results for a given file. It would be nice to display all of them, or even allow filtering on the test name!

justarandomgeek:
This fixes #245 for me.

### DIFF
--- a/@types/cobertura-parse/index.d.ts
+++ b/@types/cobertura-parse/index.d.ts
@@ -42,6 +42,7 @@ declare namespace parse {
     }
 
     interface Section {
+        title: string,
         branches?: Branches,
         file: string,
         functions: Functions,
--- a/@types/jacoco-parse/index.d.ts
+++ b/@types/jacoco-parse/index.d.ts
@@ -41,6 +41,7 @@ declare namespace parse {
     }
 
     interface Section {
+        title: string,
         branches?: Branches,
         file: string,
         functions: Functions,
--- a/@types/lcov-parse/index.d.ts
+++ b/@types/lcov-parse/index.d.ts
@@ -38,6 +38,7 @@ declare namespace parse {
     }
 
     interface Section {
+        title: string,
         branches?: Branches,
         file: string,
         functions: Functions,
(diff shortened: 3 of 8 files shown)


ryanluker:
@justarandomgeek thanks for the PR!
This is interesting, I assumed that the section returned from the different parsers (cobertura, lcov parse, clover, etc) would be unique per file and not greater then 1. Do you have an example lcov file that I can take a look at that shows this?

We should also add some integration test coverage to cover this new feature!
I can assist with a bit of the coding later next week if you dont get to fixing the linting and the additional test coverage.

justarandomgeek:
Sure, here's a file produced by my Factorio Mod coverage tool: it's separated out by test name and the Lua instance in which the script ran during that test, which i've sort of bodged together into a composite test name for lack of a better place to put the Lua instance name. The same file might sometimes run in multiple different Lua instances, and obviously different tests can go back over all the same files!

https://gist.github.com/justarandomgeek/8660e0685276994165915bb3df2ca347

Also, at least one of your own examples has this: https://github.com/ryanluker/vscode-coverage-gutters/blob/master/example/node/lcov.info has 94 sections for "SF:./node/test.js" all with blank TN

I'm afraid my js/ts skill is probably not great enough to help much though - i just barely managed to hack this patch into working at all! :)

ryanluker:
@justarandomgeek thanks for all the extra info! I should be able to take this away and finish it for yah :grin: .

**Side discussion:** It is very cool to see you use the extension for factorio mods, I have played the base game a bit and it is quite fun (seems to scratch that programming itch)!
What is the name of the mod you are making?

ryanluker:
@justarandomgeek I took your changes and moved them to a new branch here #247

justarandomgeek:
The one that brought me here specifically was my [Coverage Tester Mod](https://github.com/justarandomgeek/factorio-coverage), which i'm using to see if my tests (which will probably eventually grow into a generalized tester tool mod...) on [ConMan](https://github.com/justarandomgeek/factorio-conman) are actually complete. I've got quite a few other mods on the factorio mod portal though, and you can see a few of their names in that lcov file!

Issue #194, opened 2018-11-28
"Count" for each line invoked/executed ?


krishnakumar3044:
When I use this plug-in, I see the red/blue colors of code coverage. Is it possible to display the "number of  times" each line is invoked/executed as part of code coverage ?

Example in html format.

src/MyCode.cpp |   |   | 26.9 % | **7 / 26** | 11.5 % | 3 / 26
-- | -- | -- | -- | -- | -- | --

ryanluker:
@krishnakumar3044 Thanks for the submitting an issue. Most coverage formats do give extract information around how many times something was traversed but the most basic information that all give is simple which lines have coverage or not. The best I could standardise on (that I could see at least) was the following data structure.

https://github.com/ryanluker/vscode-coverage-gutters/blob/master/%40types/lcov-parse/index.d.ts#L4-L45

There is "hit" information that might accomplish what you want but the next step would be to flush out how this would look in the IDE (I think jet brains developer tools show this "hit" information in an on hover tooltip). Any ideas on where this information might appear?

felipecaputo:
@ryanluker maybe makes like GitLens do, with an aditional info that show side by side of the source line 

![image](https://user-images.githubusercontent.com/7751449/50520976-724edd80-0aaa-11e9-8fa9-427a08ce7ecc.png)

ryanluker:
@felipecaputo thanks for example, seems like a reasonable location for this extra info.

GitMensch:
code lens solution available with https://github.com/JacquesLucke/gcov-viewer in files using the same license - I think the code could be borrowed there

ryanluker:
> code lens solution available with https://github.com/JacquesLucke/gcov-viewer in files using the same license - I think the code could be borrowed there

Thanks for another example @GitMensch !

PR #247, opened 2019-11-27
Allow multiple sections per file path

ryanluker:
@justarandomgeek FYI
Related Issue: #245 
Related PR: #246 

## Work
- [x] Adjust section finding and rendering to operate on multiple sections
- [x] Deduplicate the coverage before creating editor decorations
- [x] Add test coverage for the scenario and or adjust the current nodejs one

### DIFF
--- a/@types/cobertura-parse/index.d.ts
+++ b/@types/cobertura-parse/index.d.ts
@@ -42,6 +42,7 @@ declare namespace parse {
     }
 
     interface Section {
+        title: string,
         branches?: Branches,
         file: string,
         functions: Functions,
--- a/@types/cobertura-parse/index.d.ts
+++ b/@types/cobertura-parse/index.d.ts
@@ -42,7 +42,7 @@ declare namespace parse {
     }
 
     interface Section {
-        title: string,
+        title?: string,
         branches?: Branches,
         file: string,
         functions: Functions,
--- a/@types/jacoco-parse/index.d.ts
+++ b/@types/jacoco-parse/index.d.ts
@@ -41,6 +41,7 @@ declare namespace parse {
     }
 
     interface Section {
+        title: string,
         branches?: Branches,
         file: string,
         functions: Functions,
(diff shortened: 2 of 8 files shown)


justarandomgeek:
I've discovered another minor issue related to this: when some tests include a DA line with 0 hits it still gets a red highlight even if another test does hit that line. In fact, it seems in general it's just stacking the highlights of all the tests currently, so i've turned the opacity down and it nicely hightlights how many tests hit a given line, but the reds ruin the effect a bit! I guess it needs to properly total up the hitcounts per line from each test instead of just stacking all of them for this to work.

ryanluker:
@justarandomgeek Ahhh interesting, with the new many sections style it must be duplicating the decorations on top of each other... Definitely a bug and I believe the logic should simply take the most "positive" result as what is showed in the editor (EG: If you have 5 coverages for a single line, 4 no coverage and 1 full coverage then the 1 full coverage should be the only one displayed).

Thoughts on this?

justarandomgeek:
Yeah, for line coverage any test hitting it should be enough to count as "full". If/when you eventually display the actual hit counts, it should be the total of all the tests that hit that line. I'm not currently generating branch coverage records (it's impractical in Lua within the restrictions of Factorio's environment), but I would think for branches you'd collect up all the branch records for a given line and combine them by the branch numbers to eventually decide on full/partial/no coverage. (I'm also generating function hit counts now, but so far only for functions that get called, and those don't seem to display anyway, which probably belongs in its own issue, if you don't already have one for that!)

ryanluker:
@justarandomgeek thanks for the extra info! All good points you made there, on the hit count not being shown... that is known future enhancement and logged here :grin: https://github.com/ryanluker/vscode-coverage-gutters/issues/194

ryanluker:
@justarandomgeek the duplicate removing can be found here https://github.com/ryanluker/vscode-coverage-gutters/pull/247/commits/d2e86a3077e96b562ab72ce859822d0daf7f9563

Thanks again for submitting the issue! I will do a release sometime within a week that will include these changes :grin: .

PR #256, opened 2020-03-16
extension was meant to be extension name.

gilescope:
At the moment the commands are under a namespace called 'extension'. That's a little confusing for anyone trying to call those commands as we'd guess vscode-coverage-gutters or coverage-gutters (I'm not entirely sure which is the most correct, but either is better than having them prefixed extension.

Figured I'd raise it as a PR rather than an issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -736,13 +736,24 @@
       }
     },
     "https-proxy-agent": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/https-proxy-agent/-/https-proxy-agent-2.2.1.tgz",
-      "integrity": "sha512-HPCTS1LW51bcyMYbxUIOO4HEOlQ1/1qRaFWcyxvwaqUS9TY88aoEuHUY33kuAh1YhVVaDQhLZsnPd+XNARWZlQ==",
+      "version": "2.2.4",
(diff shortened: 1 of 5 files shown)


gilescope:
@ryanluker thoughts? are you ok with this change? I know it's technically a breaking change for anyone that depends upon calling your actions by name, but I think it's the right thing to do.

ryanluker:
@gilescope this is looking good, thank you very much for the pull request! I will give it an official review either later this week or sometime on the weekend.

I don't mind breaking changes but as you mentioned if we wanted to be exact we would have to put this into the 3.0.0 release I have planned. I don't yet know when that release would be going out as very little work has been completed on it so far :thinking-face:.

gilescope:
Sure that’s fine. I’ll fix up my extension when it changes :-) no rush.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ryanluker:
@gilescope Thanks for updating this!
Sometime over the next couple days I am hoping to make a "beta" vsix and create a github release for this and other improvements in the meantime until 3.0.0 is further along.
Once this vsix is made you can manually install it to get your changes sooner
https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix

Issue #260, opened 2020-04-23
PHP: Non-namespaced classes do not render coverage


webmaster777:
**Describe the bug**
I have a project that contains both namespaced and non-namespaced classes. I generate coverage for both. The coverage is shown in namespaced classes, but is omitted at non-namespaced classes

**To Reproduce**
Steps to reproduce the behaviour:
1. create a project with **both** a non-namespaced ("global") and a namespaced php class
2. create coverage report with phpunit
3. open file and render coverage or enable watch

**Expected behaviour**
See coverage in gutter.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macos
 - Extension Version 2.4.3

vscode:
```
Version: 1.44.2
Commit: ff915844119ce9485abfe8aa9076ec76b5300ddd
Date: 2020-04-16T17:07:18.473Z
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Darwin x64 18.7.0
```

**Additional context (logs)**

namespaced classes:
```log
[1587629969374][coverageservice]: READY
[1587629969458][coverageservice]: RENDERING
[1587629969462][renderer][section test name]: SomeNamespaced\Thing
[1587629969463][renderer][section file path]: /path/to/my/classes/SomeNamespaced/Thing.php
[1587629969472][coverageservice]: READY
```

non-namespaced class:
```
[1587629974310][coverageservice]: RENDERING
[1587629974310][coverageservice]: READY
[1587629974423][coverageservice]: RENDERING
[1587629974432][coverageservice]: READY
[1587629987700][coverageservice]: RENDERING
[1587629987701][coverageservice]: READY
[1587630108756][coverageservice]: RENDERING
[1587630108758][coverageservice]: READY
<!-- Here I tried to load via cmd+shift+7 -->
[1587630115283][coverageservice]: LOADING
[1587630126231][coverageservice]: Loading 2 file(s)
[1587630126279][coverageservice]: Loaded 2 data file(s)
[1587630126896][coverageservice]: Caching 265 coverage(s)
[1587630126896][printDataCoverage]: Coverage -> 265
[1587630126896][coverageservice]: READY
[1587630126897][coverageservice]: RENDERING
```

**edit**
I've adapted the reproduction to explicitly mention to have both namespaced and non-namespaced classes in the same project.

ryanluker:
[...]

PR #261, opened 2020-05-06
PHP: non-namespaced example

webmaster777:
https://github.com/ryanluker/vscode-coverage-gutters/issues/260#issuecomment-618812480

### DIFF
--- a/example/php/composer.json
+++ b/example/php/composer.json
@@ -20,12 +20,12 @@
   "license": "MIT",
   "type": "library",
   "require": {
-    "php": ">=5.3.0",
-    "monolog/monolog": "1.*"
+    "php": "^7.2"
   },
   "require-dev": {
-    "phpdocumentor/phpdocumentor": "2.*",
-    "phpunit/phpunit": "4.*",
-    "sebastian/phpcpd": "2.*"
+    "phpunit/phpunit": "^8"
+  },
+  "autoload": {
+    "classmap": ["src/"]
   }
 }
--- a/package.json
+++ b/package.json
@@ -270,7 +270,7 @@
     "vscode": "1.1.34"
   },
   "dependencies": {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -270,7 +270,7 @@
     "vscode": "1.1.34"
   },
   "dependencies": {
-    "@cvrg-report/clover-json": "0.3.0",
+    "@cvrg-report/clover-json": "^0.3.2",
```
ryanluker, line 273: can you leave this pinned to `0.3.2`?

test/extension.test.ts
```diff
@@ -137,6 +137,28 @@ suite("Extension Tests", function() {
         extension.exports.emptyLastCoverage();
     });
 
+    test("Run display coverage on php test file number 3 @integration", async () => {
```
ryanluker, line 140: excellent thanks a bunch for adding a test to cover this case!

example/php/composer.json
```diff
@@ -20,12 +20,12 @@
   "license": "MIT",
   "type": "library",
   "require": {
-    "php": ">=5.3.0",
-    "monolog/monolog": "1.*"
+    "php": "^7.2"
```
ryanluker, line 23: thanks for updating this and all the other php files! (I have very little to no experience with it so this is super helpful).

package.json
webmaster777: Sure, though I can't imagine why you want to pin all these dependencies. Bad experience with B/C breaks between minor versions?

ryanluker: haha yep... normally in most software projects I let these float atleast to the minor / patch but freeze the dependencies for builds via the package-lock.json

webmaster777:
> Thanks again! I will try to get the release to the marketplace sometime this week.

Any chance you can create a new patch release? If there's anything I can do to help the release process, let me know!

PR #271, opened 2020-07-16
Make base glob configurable via coverageBaseDir

amatosov-rbx:
# Rational

On big workspaces (e.g. 12k directories, 74k files) recursive glob lookup `**` is very time consuming. Currently there is no way to avoid it via configuration options. 
`manualCoverageFilePaths` is the only option, however it eliminates any kind of glob lookup, which is not desired.

# Change

Introduce `coverageBaseDir` that allows users to fully customize lookup glob for coverage files
It defaults to `**`, so existing behavior is not going to change.

### DIFF
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -135,6 +135,8 @@ export class CoverageService {
         const fileNames = this.configStore.coverageFileNames.toString();
         let baseDir = this.configStore.coverageBaseDir;
         if (workspace.workspaceFolders) {
+            // Prepend workspace folders glob to the folder lookup glob
+            // This allows watching within all the workspace folders
             const workspaceFolders = workspace.workspaceFolders.map((wf) => wf.uri.fsPath);
             baseDir = `{${workspaceFolders}}/${baseDir}`;
         }
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -140,7 +140,7 @@ export class CoverageService {
         }
 
         // Creates a BlobPattern for all coverage files.
-        // EX: `**/{cov.xml, lcov.info}`
+        // EX: `{/path/to/workspace1, /path/to/workspace2}/**/{cov.xml, lcov.info}`
         const blobPattern = `${baseDir}/{${fileNames}}`;
         this.coverageWatcher = workspace.createFileSystemWatcher(blobPattern);
         this.coverageWatcher.onDidChange(this.loadCacheAndRender.bind(this));
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -133,9 +133,15 @@ export class CoverageService {
         if (this.configStore.manualCoverageFilePaths.length) { return; }
 
         const fileNames = this.configStore.coverageFileNames.toString();
+        let baseDir = this.configStore.coverageBaseDir;
+        if (workspace.workspaceFolders) {
+            const workspaceFolders = workspace.workspaceFolders.map((wf) => wf.uri.fsPath);
+            baseDir = `{${workspaceFolders}}/${baseDir}`;
+        }
+
         // Creates a BlobPattern for all coverage files.
         // EX: `**/{cov.xml, lcov.info}`
-        const blobPattern = `**/{${fileNames}}`;
+        const blobPattern = `${baseDir}/{${fileNames}}`;
         this.coverageWatcher = workspace.createFileSystemWatcher(blobPattern);
         this.coverageWatcher.onDidChange(this.loadCacheAndRender.bind(this));
         this.coverageWatcher.onDidCreate(this.loadCacheAndRender.bind(this));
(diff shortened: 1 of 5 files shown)


src/coverage-system/coverageservice.ts
```diff
@@ -133,9 +133,15 @@ export class CoverageService {
...
 
         const fileNames = this.configStore.coverageFileNames.toString();
+        let baseDir = this.configStore.coverageBaseDir;
+        if (workspace.workspaceFolders) {
+            const workspaceFolders = workspace.workspaceFolders.map((wf) => wf.uri.fsPath);
+            baseDir = `{${workspaceFolders}}/${baseDir}`;
+        }
+
         // Creates a BlobPattern for all coverage files.
         // EX: `**/{cov.xml, lcov.info}`
```
ryanluker, line 143: maybe update this comment to include an example for the new `coverageBaseDir`

src/coverage-system/coverageservice.ts
```diff
@@ -133,9 +133,15 @@ export class CoverageService {
...
-        const blobPattern = `**/{${fileNames}}`;
+        const blobPattern = `${baseDir}/{${fileNames}}`;
```
ryanluker, line 144: This will effect your other PR as well but I am guessing you have tested these together?
Although by my reading it should be mutually exclusive in that you can have either this blobPattern that has quite a long of flexibility or you can do the manual approach?

src/coverage-system/coverageservice.ts
```diff
@@ -133,9 +133,15 @@ export class CoverageService {
         if (this.configStore.manualCoverageFilePaths.length) { return; }
 
         const fileNames = this.configStore.coverageFileNames.toString();
+        let baseDir = this.configStore.coverageBaseDir;
+        if (workspace.workspaceFolders) {
+            const workspaceFolders = workspace.workspaceFolders.map((wf) => wf.uri.fsPath);
+            baseDir = `{${workspaceFolders}}/${baseDir}`;
```
ryanluker, line 139: so the full glob will be as follows if I am understanding this correctly...
`{c/workspacefolderc/coverages, b/workspacefolderb/coverages, a/workspacefoldera/coverages,}`
and then when combined below the blob is:
`{c/workspacefolderc/coverages, b/workspacefolderb/coverages, a/workspacefoldera/coverages,}/{cov.xml, lcov.info}`

If that is correct then I think that works here! as per the other PR you may want to write a few tests to protect this functionality into the future 🤔.

Integration while best might need you to make a new example folder like the remote-node one.. .`example\remote-node`

src/files/filesloader.ts
```diff
@@ -84,7 +84,7 @@ export class FilesLoader {
         fileName: string,
     ) {
         return new Promise<Set<string>>((resolve, reject) => {
-            glob(`**/${fileName}`,
+            glob(`${this.configStore.coverageBaseDir}/${fileName}`,
```
ryanluker, line 87: do we need to consider the workspace folder as apart of this blob or are we hoping the cwd will assist us with that?

amatosov-rbx: `cwd` is sufficient in this case as this glob lookup done per workspace folder. It is possible to create combined glob pattern, same as in `coverageservice.ts` but is more intrusive change.
From end user prospective they yield the same result.

src/coverage-system/coverageservice.ts
amatosov-rbx: Yes, that is the resulting pattern. I'll add tests and examples

src/coverage-system/coverageservice.ts
amatosov-rbx: 1. Yes, they are mutually exclusive
1. Yes I have tested them together (and separately)
1. They will conflict in this file, after one of them is merged, so I'll have to update second review
1. This change technically gives enough flexibility for glob setup of any complexity. manual approach can still be used for simple cases where you have just one or two explicit lcov files you want to watch

amatosov-rbx:
Converted to draft. #270 should go in first

src/coverage-system/coverageservice.ts
amatosov-rbx: done

src/coverage-system/coverageservice.ts
amatosov-rbx: I have added unit tests, however not sure what to add to examples. They are used in integration tests, but testing watch changes in-tree is problematic (as described in https://github.com/ryanluker/vscode-coverage-gutters/pull/270#issuecomment-660368276)

amatosov-rbx:
@ryanluker this PR is ready for review

amatosov-rbx:
@ryanluker Do you have any updates on the 2.6.0 release date?

ryanluker:
@amatosov-rbx I was hoping the last item from the milestone would get in but if it isn't by Sunday I will just go away with a release. https://github.com/ryanluker/vscode-coverage-gutters/milestone/31

ryanluker:
@amatosov-rbx 2.6.0 should be out now! thanks again for the effort in adding / fixing this 😁 .
https://github.com/ryanluker/vscode-coverage-gutters/releases/tag/v2.6.0

amatosov-rbx:
@ryanluker Thanks for getting the release out and quick feedback on fixes and additions =)

PR #290, opened 2021-01-29
refactor: Webview creation logic in new module

Kaytbode:
Webview.ts file created in the extension directory, and a
CSP header is being added to text before loading it to webview.html.

### DIFF
--- a/src/extension/gutters.ts
+++ b/src/extension/gutters.ts
@@ -1,15 +1,13 @@
 import {
     OutputChannel,
-    Uri,
-    ViewColumn,
     window,
-    workspace,
 } from "vscode";
 import { Coverage } from "../coverage-system/coverage";
 import { CoverageService } from "../coverage-system/coverageservice";
 import { Config } from "./config";
 import { CrashReporter } from "./crashreporter";
 import { StatusBarToggler } from "./statusbartoggler";
+import { PreviewPanel } from "./webview";
 export class Gutters {
     private coverage: Coverage;
     private outputChannel: OutputChannel;
@@ -42,17 +40,9 @@ export class Gutters {
             );
             if (!pickedReport) { throw new Error("Could not show Coverage Report file!"); }
 
-            // Construct the webview panel for the coverage report to live in
-            const previewPanel = window.createWebviewPanel(
-                "coverageReportPreview",
-                "Preview Coverage Report",
-                ViewColumn.One,
-            );
+            const previewPanel = new PreviewPanel(pickedReport);
 
-            // Read in the report html and send it to the webview
-            const reportUri = Uri.file(pickedReport);
-            const reportHtml = await workspace.openTextDocument(reportUri);
-            previewPanel.webview.html = reportHtml.getText();
+            await previewPanel.createWebView();
         } catch (error) {
             this.handleError("previewCoverageReport", error);
         }
--- a/src/extension/webview.ts
+++ b/src/extension/webview.ts
@@ -12,12 +12,6 @@ export class PreviewPanel {
 
     constructor(pickedReport: string) {
         this.pickedReport = pickedReport;
-        // Construct the webview panel for the coverage report to live in
-        this.previewPanel = window.createWebviewPanel(
-            "coverageReportPreview",
-            "Preview Coverage Report",
-            ViewColumn.One,
-        );
     }
 
     public async createWebView() {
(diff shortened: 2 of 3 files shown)


src/extension/webview.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.previewPanel.webview.html = reportHtmlWithPolicy;
+    }
+
+    private addContentSecurityPolicy(text: string) {
+        const meta = text.indexOf("<meta");
+        const securityPolicyHeader = `<meta http-equiv="Content-Security-Policy" content="default-src 'none';">`;
+
+        const newText = `${text.substring(0, meta)}${securityPolicyHeader}${text.substring(meta)}`;
+
+        return newText;
```
ryanluker, line 38: nice work injecting this into the html!

src/extension/webview.ts
```diff
@@ -0,0 +1,40 @@
...
+    public async createWebView() {
+        // Read in the report html and send it to the webview
+        const reportUri = Uri.file(this.pickedReport);
+        const reportHtml = await workspace.openTextDocument(reportUri);
+        const reportHtmlWithPolicy = this.addContentSecurityPolicy(reportHtml.getText());
+
+        this.previewPanel.webview.html = reportHtmlWithPolicy;
+    }
+
+    private addContentSecurityPolicy(text: string) {
```
ryanluker, line 32: minor: we could add a return type of string here I guess 🤔

src/extension/gutters.ts
```diff
@@ -42,17 +40,9 @@ export class Gutters {
...
-                "Preview Coverage Report",
-                ViewColumn.One,
-            );
+            const previewPanel = new PreviewPanel(pickedReport);
 
-            // Read in the report html and send it to the webview
-            const reportUri = Uri.file(pickedReport);
-            const reportHtml = await workspace.openTextDocument(reportUri);
-            previewPanel.webview.html = reportHtml.getText();
+            await previewPanel.createWebView();
```
ryanluker, line 45: ### Discussion
We might have to store the preview panel on the instance so we can can call dispose later (`public dispose()`).
I have a bit of pseudo code below, let me know what you think.

### Research
https://code.visualstudio.com/api/references/vscode-api#WebviewPanel

### Pseudo Code
```typescript
this.previewPanel = new PreviewPanel(pickedReport);

~~~~

public dispose() {
    this.previewPanel.dispose();
}
```

ryanluker (approved):
Looking great! Just a couple minor comments and a couple discussion points.

### Regarding Testing:
I believe you also have the testing working that we have for this (just integration ones really) but we could also add a simple unit style one to make sure the `addContentSecurityPolicy` properly appends the policy. We could also consider then what happens when it can't find the meta tag (does it gracefully fail?)

src/extension/webview.ts
Kaytbode: Ok. will do that

src/extension/webview.ts
Kaytbode: Ok. Will do that

src/extension/webview.ts
Kaytbode: Thank you

src/extension/gutters.ts
Kaytbode: > Looking great! Just a couple minor comments and a couple discussion points.
> 
> ### Regarding Testing:
> I believe you also have the testing working that we have for this (just integration ones really) but we could also add a simple unit style one to make sure the `addContentSecurityPolicy` properly appends the policy. We could also consider then what happens when it can't find the meta tag (does it gracefully fail?)

@ryanluker Yes, the integration test is working. I will work on writing the unit tests.

Kaytbode: @ryanluker this is how I want to go about writing the unit test to check if `addContentSecurityPolicy` properly appends the policy. 

1. Create a variable that holds an HTML text.
2. Make the `addContentSecurityPolicy` property of PreviewPanel public, so that I can have access to it in the test.
3. Check if the returned value from calling `addContentSecurityPolicy` contains the content security policy header.

What do you think about my approach? I also think it is more logical to check for the presence of a _head_ tag than a _meta_ tag?

ryanluker: That approach sounds great @Kaytbode ! you could also add a failing unit test to confirm what we expect to happen in that scenario 🤔 .

In terms of approach, I am not sure which is more common in these preview reports (we only have the node generated one so there could be some variability). We could also support both options (check for meta, if not found try head but still add just one security policy).

Kaytbode: @ryanluker You are yet to review the changes I made

ryanluker: @Kaytbode Thanks for the reminder, I will review again shortly.

src/extension/webview.ts
```diff
@@ -0,0 +1,49 @@
...
+        );
+
+        this.previewPanel.webview.html = reportHtmlWithPolicy;
+    }
+
+    public dispose() {
+        this.previewPanel.dispose();
+    }
+
+    public addContentSecurityPolicy(text: string): string {
```
ryanluker, line 37: this looks like it is only used by `PreviewPanel` maybe we should make this private?

ryanluker (approved):
Looking great @Kaytbode , just one minor comment. Thanks for adding tests as well!

src/extension/webview.ts
Kaytbode: @ryanluker Thank you for the review. I initially set the `method` to be private, but the tests I wrote, require I make it public. Please I will need your advice on how to go about it

ryanluker: I see! Lets leave it public then and merge this into the mainline.

Kaytbode: @ryanluker pleasure is mine... Thanks for being a patient guide

Kaytbode: @ryanluker waiting on this merge so I can take on a new issue

ryanluker: Oh it does look like I need to merge this manually then... makes sense!

PR #297, opened 2021-03-18
Allow named workspace folders

### LINKED ISSUE #252: Coverage not displaying in Workspace
**Describe the bug**
I have a code-workspace file setup with multiple root folders. When I open just the folder with the source I see the coverage just fine. when I use the code-workspace it doesn't display the coverage and just reports this in the output tab:
```
[1580939108663][coverageservice]: INITIALIZING
[1580939870656][coverageservice]: LOADING
[1580939873483][coverageservice]: Loading 1 file(s)
[1580939873484][coverageservice]: Loaded 1 data file(s)
[1580939873492][coverageservice]: Caching 51 coverage(s)
[1580939873492][printDataCoverage]: Coverage -> 51
[1580939873492][coverageservice]: READY
[1580939873492][coverageservice]: RENDERING
[1580939873493][coverageservice]: READY
[1580939874648][coverageservice]: RENDERING
[1580939874649][coverageservice]: READY
```

**Expected behaviour**
Code coverage should highlight the lines in the files.

**Desktop (please complete the following information):**
 - OS: Windows
 - Extension Version: 2.4.3
 - VSCode Version: 1.41.1

evan-bradley:
This PR fixes #252.

The change is to use the base name of the path for the workspace folder instead of the name, which defaults to the path's base name if a name hasn't been specified. I've tested that this works in my setup (VS Code 1.54.2 running on Linux), and have added unit tests to cover the cases where a name has or hasn't been specified.

### DIFF
--- a/src/coverage-system/sectionfinder.ts
+++ b/src/coverage-system/sectionfinder.ts
@@ -1,5 +1,5 @@
 import {Section} from "lcov-parse";
-import {extname} from "path";
+import {basename} from "path";
 import {TextEditor, Uri, workspace} from "vscode";
 import {OutputChannel} from "vscode";
 import {Config} from "../extension/config";
@@ -116,7 +116,7 @@ export class SectionFinder {
         const editorFileAbs = normalizeFileName(fileName);
         const workspaceFile = normalizeFileName(workspaceFsPath);
         const editorFileRelative = editorFileAbs.substring(workspaceFile.length);
-        const workspaceFolderName = normalizeFileName(workspaceFolder.name);
+        const workspaceFolderName = normalizeFileName(basename(workspaceFsPath));
         return { relativePath: editorFileRelative, workspaceFolder: workspaceFolderName};
     }
 
--- a/test/coverage-system/sectionfinder.test.ts
+++ b/test/coverage-system/sectionfinder.test.ts
@@ -1,10 +1,17 @@
-import {Section} from "lcov-parse";
-import {TextEditor} from "vscode";
+import * as assert from "assert";
+import { Section } from "lcov-parse";
+import { basename, join } from "path";
+import { TextEditor, Uri, workspace, WorkspaceFolder } from "vscode";
 
-import {SectionFinder} from "../../src/coverage-system/sectionfinder";
-import {fakeConfig} from "../mocks/fakeConfig";
+import { SectionFinder } from "../../src/coverage-system/sectionfinder";
+import { fakeConfig } from "../mocks/fakeConfig";
+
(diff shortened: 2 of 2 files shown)


src/coverage-system/sectionfinder.ts
```diff
@@ -116,7 +116,7 @@ export class SectionFinder {
         const editorFileAbs = normalizeFileName(fileName);
         const workspaceFile = normalizeFileName(workspaceFsPath);
         const editorFileRelative = editorFileAbs.substring(workspaceFile.length);
-        const workspaceFolderName = normalizeFileName(workspaceFolder.name);
+        const workspaceFolderName = normalizeFileName(basename(workspaceFsPath));
```
ryanluker, line 119: Looking pretty good, I can test on windows once this is in master (and the integration tests pass so probably not an issue). I did see that the windows usage might be different (1) but hopefully the integration tests catch this.

1: https://nodejs.org/api/path.html#path_path_basename_path_ext

src/coverage-system/sectionfinder.ts
```diff
@@ -1,5 +1,5 @@
 import {Section} from "lcov-parse";
-import {extname} from "path";
+import {basename} from "path";
```
ryanluker, line 2: we were evening usign extname before?

ryanluker (approved):
Looking great mostly a couple questions for my understanding

ryanluker:
@evan-bradley Going to merge this one into the mainline and the changes will go out with the next release (2.7.3)
I left a few comments that are still outstanding but they were mainly questions for my own understanding.

src/coverage-system/sectionfinder.ts
evan-bradley: Hey, apologies for not getting back to you on this. This shouldn't be an issue since the case-sensitivity only affects file extensions from what I can tell in the documentation. VS Code doesn't support pointing workspace folders to files, so this shouldn't be an issue.

src/coverage-system/sectionfinder.ts
evan-bradley: It wasn't being used according to VS Code and a quick search in the file.

src/coverage-system/sectionfinder.ts
ryanluker: No problem on not getting back right away, sounds good on the reasoning as well.

src/coverage-system/sectionfinder.ts
ryanluker: alright cool!

Issue #239, opened 2019-10-20
Color settings don't work


tiffany352:
**Describe the bug**
I'm red-green colorblind (the most common type, which affects 1 in 25 people). This app uses shades of red and green of very similar lightness and hue which I have to squint to tell apart.

I see that there's settings for "Highlightdark" and "Highlightlight" but if I change them then nothing seems to happen. It seems like I have to edit those SVGs that are hardcoded to red and green in order for my changes to have an effect...

**To Reproduce**
Steps to reproduce the behaviour:
1. Be colorblind
2. Try to tell apart the red and green gutter colors
3. Try to change settings to fix it
4. The settings have no effect

**Expected behaviour**
Ideally the extension just wouldn't use red and green by default. It would also be fine if there was an easy way to use an alternate color scheme though.

**Desktop (please complete the following information):**
 - OS: Windows 10 (using remote SSH extension though, so this extension is installed on a Fedora Server 29 host)
 - Extension Version: 2.4.1
 - VSCode Version: 1.39.2

ryanluker:
@tiffany352 Thanks for this issue as well. You are correct that there is no easy way to swap to a colour blind mode (via a single setting) but you can modify the colours of both the gutter and line coverage.

1) To modify the line coverage you can simply adjust the rgba to your liking of the theme you are using (looks to be dark from your other issue). https://github.com/ryanluker/vscode-coverage-gutters/blob/master/package.json#L47-L76

2) To adjust the gutters, you are correct that they need to be svgs, you will need to provide a substitute svg file path in the `settings.json` to use instead. https://github.com/ryanluker/vscode-coverage-gutters/blob/master/package.json#L77-L86. See my screenshare below where I modify the svg to a different hue (I have also attached the 3 svg files so you can easily modify the hues).

Let me know if that helps any, I will add a new issue though to make a colour blind friendly setting that will manually force different hues (any suggestions on what the three colours should be?)

[icons-dark.zip](https://github.com/ryanluker/vscode-coverage-gutters/files/3748320/icons-dark.zip)
https://imgur.com/9atua5J

tiffany352:
I tried replacing the SVG but it didn't seem to work either. I'm not sure what I'm doing wrong with it. I currently only have coverage files in a repo that's on remote SSH, so I can't really tell if there's something specific to that I'm doing wrong. I tried putting the file both on my local PC and on the remote host and using the path to it. In both cases, the gutter image doesn't render.

[...]

PR #310, opened 2021-05-22
Enable custom svgs to be provided via config options

mattseddon:
I ran into an issue today whilst trying to update the colors of gutters' SVGs (same issue as detailed in #239).

These changes are designed to enable users to have permanently saved SVGs somewhere on disk.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,32 +76,26 @@
         },
         "coverage-gutters.gutterIconPathLight": {
           "type": "string",
-          "default": "./app_images/gutter-icon-light.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for full coverage"
         },
         "coverage-gutters.gutterIconPathDark": {
           "type": "string",
-          "default": "./app_images/gutter-icon-dark.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for full coverage"
         },
         "coverage-gutters.partialGutterIconPathLight": {
           "type": "string",
-          "default": "./app_images/partial-gutter-icon-light.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for partial coverage"
         },
         "coverage-gutters.partialGutterIconPathDark": {
           "type": "string",
-          "default": "./app_images/partial-gutter-icon-dark.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for partial coverage"
         },
         "coverage-gutters.noGutterIconPathLight": {
           "type": "string",
-          "default": "./app_images/no-gutter-icon-light.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for no coverage"
         },
         "coverage-gutters.noGutterIconPathDark": {
           "type": "string",
-          "default": "./app_images/no-gutter-icon-dark.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for no coverage"
         },
         "coverage-gutters.showLineCoverage": {
--- a/src/extension/config.ts
+++ b/src/extension/config.ts
@@ -70,27 +70,34 @@ export class Config {
         const partialCoverageDarkBackgroundColour = rootConfig.get("partialHighlightDark") as string;
         const noCoverageLightBackgroundColour = rootConfig.get("noHighlightLight") as string;
         const noCoverageDarkBackgroundColour = rootConfig.get("noHighlightDark") as string;
-        const gutterIconPathDark = rootConfig.get("gutterIconPathDark") as string;
-        const gutterIconPathLight = rootConfig.get("gutterIconPathLight") as string;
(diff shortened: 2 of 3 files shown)


ryanluker:
@mattseddon Thanks for the contribution!
I will try to take a look at it over the weekend sometime. I have authorized builds in the meantime though so that you can take a look at the linting errors etc.

mattseddon:
> @mattseddon Thanks for the contribution!
> I will try to take a look at it over the weekend sometime. I have authorized builds in the meantime though so that you can take a look at the linting errors etc.

No worries, thought that I had checked that the linter passed before putting up the PR but obviously not. I would be happy to get a `husky` pre-commit hook running for you if that is something that you think would be valuable. I would move your tests to use the insiders build at the same time because then you can run the tests whilst developing inside of vscode.

Anyway, LMK if you think that would be valuable. If it is I'll raise a separate PR.

ryanluker:
@mattseddon great thinking on both fronts!
**RE pre-commit**: Good idea, I have utilized pre-commit before (https://pre-commit.com/) for some work python projects. Based on a quick google it appears husky is similar (https://github.com/typicode/husky)? If so it sounds like a good addition, I would also have it added to the dev container https://github.com/ryanluker/vscode-coverage-gutters/blob/master/.devcontainer/devcontainer.json#L24-L25 and maybe a mention in the readme around contributing 🤔 

**RE insiders build:** That was also on my mental backlog so that sounds great to me!

test/extension/config.test.ts
```diff
@@ -74,12 +84,25 @@ suite("Config Tests", function() {
...
+        showGutterCoverage = true;
+        iconPathDark = "/my/absolute/path/to/custom/icon-dark.svg";
+        iconPathLight = "";
+        (vscode as any).window.createTextEditorDecorationType = (options) => {
+            assert.equal(options.dark.gutterIconPath, iconPathDark);
+            assert.equal(options.light.gutterIconPath.includes("./app_images/"), true);
+        };
+        fakeContext.asAbsolutePath = (options) => options;
+        new Config(fakeContext); // tslint:disable-line
+    });
```
ryanluker, line 106: nice test additions!

src/extension/config.ts
```diff
@@ -70,27 +70,34 @@ export class Config {
...
+            gutterIconPathLight: "./app_images/gutter-icon-light.svg",
+            noGutterIconPathDark: "./app_images/no-gutter-icon-dark.svg",
+            noGutterIconPathLight: "./app_images/no-gutter-icon-light.svg",
+            partialGutterIconPathDark: "./app_images/partial-gutter-icon-dark.svg",
+            partialGutterIconPathLight: "./app_images/partial-gutter-icon-light.svg",
+        };
+
+        const getIcon = (name: string): string =>
+            rootConfig.get(name) as string ||
+            this.context.asAbsolutePath(defaultIcons[name]);
```
ryanluker, line 88: nice abstraction / single liner!

package.json
```diff
@@ -76,32 +76,26 @@
...
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for partial coverage"
         },
         "coverage-gutters.noGutterIconPathLight": {
           "type": "string",
-          "default": "./app_images/no-gutter-icon-light.svg",
           "description": "path to an icon (svg, png, etc) for displaying in the gutter for no coverage"
         },
         "coverage-gutters.noGutterIconPathDark": {
           "type": "string",
-          "default": "./app_images/no-gutter-icon-dark.svg",
```
ryanluker, line 104: I dont believe moving the defaults for these icon paths into the source code will force us to do a breaking version change 🤔.
It will make it appear like you dont get a default in the config settings I think. I don't believe many will notice the change though.
image

ryanluker (approved):
Just a couple comments and a discussion item on removing the default from some of the settings. Looks good to me otherwise and can be merged whenever 😁 .

**Note: ** I added the milestone 2.7.4 to this PR and the github ticket as it will go out in the next minor release.

package.json
mattseddon: I built the `vsix` and installed locally to have a look / double check, when no default is provided this is what is shown on screen (see `coverage-gutters.gutterIconPathLight`:

image

From the `WorkspaceConfiguration` API:

```
		/**
		 * Return a value from this configuration.
		 *
		 * @param section Configuration name, supports _dotted_ names.
		 * @param defaultValue A value should be returned when no value could be found, is `undefined`.
		 * @return The value `section` denotes or the default.
		 */
		get<T>(section: string, defaultValue: T): T;
```

If no value is found `undefined` will be returned and the code will work as expected so we can get away without a default.

The potential breaking change will be to users who have figured out that they can save their own custom SVGs with a path relative to where the extension is installed. I don't see that as much of an issue as I think only a handful of people would have worked it out.

Potential we could change the wording in the setting to `path to a custom icon (svg, png, etc) for displaying in the gutter...` which would implicitly let people know that there is a default.

LMK what you think.

Thanks,

PR #311, opened 2021-05-24
Use insiders build for test suite

mattseddon:
As discussed here: https://github.com/ryanluker/vscode-coverage-gutters/pull/310#issuecomment-846496483.

Change means that you can run the test suite from the command line whilst `vscode` is open. 👍🏻


ryanluker (approved):
@mattseddon Thanks for adjusting this!

PR #314, opened 2021-05-29
Add husky pre-commit hook

mattseddon:
This PR adds a [Husky](https://typicode.github.io/husky/#/) as promised [here](https://github.com/ryanluker/vscode-coverage-gutters/pull/310#issuecomment-846496483). 

The hook runs `tslint --fix` against all staged files before allowing a commit to be processed (if `tslint` cannot fix the issues in the staged file then the commit will not proceed). 

I initially could not get tests to run from the command line inside of the `devcontainer`. There are further changes required to get that part of the hook to run and I felt that stopping here and asking the question as to whether or not it is desirable to run the tests in the pre-commit hook would be the best way to proceed.

I looked into the `contributing.md` and did not see any need to mention the new hook. I have however added the prepare script to the devcontainer so that the hooks are installed after it is built. Again, happy to change if you see things differently 👍🏻.

LMK if you would like to see any changes or additions to this PR 👍🏻. Happy to raise a follow up to add the tests in, if desired.


package.json
```diff
@@ -245,13 +245,18 @@
...
     "test-integration": "cross-env MOCHA_GREP=@integration npm test --silent",
     "vscode:prepublish": "tsc -p ./",
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
-    "build": "npm install && npm run lint && npm run compile"
+    "build": "npm install && npm run lint && npm run compile",
+    "prepare": "husky install"
+  },
+  "lint-staged": {
+	"{src,test}/**/*.ts": [ "tslint --fix" ]
```
mattseddon, line 259: [F] Lifted from https://github.com/typicode/husky/issues/949#issuecomment-821917498

.devcontainer/devcontainer.json
```diff
@@ -22,8 +22,8 @@
 	"forwardPorts": [3000],
 
 	// Specifies a command that should be run after the container has been created.
-	"postCreateCommand": "npm install",
+	"postCreateCommand": "npm install && npm run prepare",
```
mattseddon, line 25: [F] For v6 of husky we have to run an extra script. Husky dropped support for autoinstall when they moved to v6. Reasoning is detailed [here](https://blog.typicode.com/husky-git-hooks-autoinstall/).

mattseddon:
I have managed to get the tests running headless from the command line inside the container using the [recommendations for setting up a Gitlab CI](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#gitlab-ci). I am pretty sure that I could get test running in the pre-commit hook for both container and non-container dev. 

@ryanluker WDYT? Is there enough value in me adding the test suite into the pre-commit hook?

.devcontainer/devcontainer.json
ryanluker: Gotcha that makes sense!

package.json
ryanluker: makes sense, nice hook to get this into everyones commit flow.

package.json
```diff
@@ -264,6 +269,8 @@
     "@types/vscode": "1.27.0",
     "chai": "4.2.0",
     "cross-env": "5.2.0",
+    "husky": "^6.0.0",
+    "lint-staged": "^11.0.0",
```
ryanluker, line 273: super minor: pin these dependencies to avoid changes (similar to the others).
I was also consider unpinning these (using the `^`) now that the CICD system is working nicely 🤔 .

ryanluker (approved):
Looking great I think we can merge this in whenever (minus the discussion on the tests).

ryanluker:
@mattseddon Thanks for the PR on the pre-commit hook!
I am not sure if we want the tests to run with each commit as that might prove pretty long time wise eventually and reduce the developer feedback loop. I think the CICD run is good enough when someone opens up a PR. I also had to disable the tests for the linux CI pipeline as it had issues running in github actions 🤔. Your vscode link to the CI documentation might give us some insight into if we can re-enable linux CI testing though.

https://code.visualstudio.com/api/working-with-extensions/continuous-integration#gitlab-ci
https://github.com/ryanluker/vscode-coverage-gutters/runs/2214395226?check_suite_focus=true

mattseddon:
> @mattseddon Thanks for the PR on the pre-commit hook!
> I am not sure if we want the tests to run with each commit as that might prove pretty long time wise eventually and reduce the developer feedback loop. I think the CICD run is good enough when someone opens up a PR. I also had to disable the tests for the linux CI pipeline as it had issues running in github actions 🤔. Your vscode link to the CI documentation might give us some insight into if we can re-enable linux CI testing though.
> 
> https://code.visualstudio.com/api/working-with-extensions/continuous-integration#gitlab-ci
> https://github.com/ryanluker/vscode-coverage-gutters/runs/2214395226?check_suite_focus=true

@ryanluker no worries, I am happy to leave tests out of the hook. 

I have pinned the new dependencies as requested.

I'll look into fixing up the linux test suite (devcontainer / CI/CD pipeline) next weekend 👍🏻.

ryanluker:
@mattseddon Thanks a ton for the contributions!

ryanluker:
@mattseddon Just used husky locally for a PR I am working on went well!
I needed to do a clean npm install though as husky tried to run in the prepare but it wasn't installed yet 🤔 .

mattseddon:
> @mattseddon Just used husky locally for a PR I am working on went well!
> I needed to do a clean npm install though as husky tried to run in the prepare but it wasn't installed yet 🤔 .

😢 that's always the way with adding new dependencies. Sorry.

ryanluker:
@mattseddon haha no need to feel sorry, I have had this happen at work a bit when adding new pre-commit type stuff so I was just wondering if it was expected.

PR #320, opened 2021-06-13
Issue 289: Add line coverage to status bar

### LINKED ISSUE #289: Code coverage percentage in bottom status bar on file change
### Description
We will need to provide the renderer with the statusbar instance so it can properly update the coverage percentage.
https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/coverage-system/coverageservice.ts#L54-L57

Then we can calculate the coverage for the "active" editor (this is to account for multiple editor panes being open) when we load the coverage lines.
https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/coverage-system/renderer.ts#L33-L65

Maybe we could have two states
Not watching 
> Screen Shot 2021-01-23 at 9 07 05 PM
Watching
> Screen Shot 2021-01-23 at 9 07 08 PM

mattseddon:
Closes #289.

~Feature works as expected but needs polished.~

@ryanluker thanks for having a look at the code whilst it was still in draft, I have refactored now and added an additional touch to alter the status bar tooltip text depending on the state of the workspace (watched vs unwatched).

Here is a v short demo (as it had to be under 10MB):

https://user-images.githubusercontent.com/37993418/121833778-66097a00-cd10-11eb-8be1-ad710bfc284e.mov

LMK if you'd like to see any changes.

Thanks,

### DIFF
--- a/package.json
+++ b/package.json
@@ -270,6 +270,7 @@
     "@types/mocha": "5.2.6",
     "@types/node": "10.14.6",
     "@types/request": "2.48.1",
+    "@types/sinon": "^10.0.2",
     "@types/uuid": "8.0.0",
     "@types/vscode": "1.27.0",
     "chai": "4.2.0",
@@ -277,6 +278,7 @@
     "husky": "6.0.0",
     "lint-staged": "11.0.0",
     "mocha": "6.1.4",
+    "sinon": "^11.1.1",
     "tslint": "5.16.0",
     "typescript": "4.2.3",
     "vscode-test": "1.5.2"
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -67,26 +67,23 @@ export class CoverageService {
         if (this.coverageWatcher) { this.coverageWatcher.dispose(); }
         if (this.editorWatcher) { this.editorWatcher.dispose(); }
         this.cache = new Map(); // reset cache to empty
-        const visibleEditors = window.visibleTextEditors;
-        this.renderer.renderCoverage(this.cache, visibleEditors);
+        this.renderer.renderCoverage(this.cache, window.visibleTextEditors);
     }
 
     public async displayForFile() {
-        await this.loadCacheAndRender();
+        await this.loadCacheAndProcess();
     }
 
     public async watchWorkspace() {
         await this.displayForFile();
         this.listenToFileSystem();
         this.listenToEditorEvents();
-        this.handleActiveEditorEvent(window.activeTextEditor);
     }
 
     public async removeCoverageForCurrentEditor() {
         try {
             this.statusBar.setLoading(true);
-            const visibleEditors = window.visibleTextEditors;
-            this.renderer.renderCoverage(new Map(), visibleEditors);
+            this.renderer.renderCoverage(new Map(), window.visibleTextEditors);
         } finally {
             this.statusBar.setLoading(false);
         }
@@ -115,13 +112,13 @@ export class CoverageService {
             `[${Date.now()}][coverageservice]: ${state}`);
     }
 
-    private async loadCacheAndRender() {
+    private async loadCacheAndProcess() {
         try {
             this.statusBar.setLoading(true);
             await this.loadCache();
             this.updateServiceState(Status.rendering);
-            const visibleEditors = window.visibleTextEditors;
-            this.renderer.renderCoverage(this.cache, visibleEditors);
+            this.renderer.renderCoverage(this.cache, window.visibleTextEditors);
+            this.setStatusBarCoverage(this.cache, window.activeTextEditor);
             this.updateServiceState(Status.ready);
         } finally {
             this.statusBar.setLoading(false);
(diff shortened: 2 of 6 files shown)


ryanluker (commented):
Took an initial look for yah but looking great so far!

package.json
```diff
@@ -270,13 +270,15 @@
...
     "@types/request": "2.48.1",
+    "@types/sinon": "^10.0.2",
     "@types/uuid": "8.0.0",
     "@types/vscode": "1.27.0",
     "chai": "4.2.0",
     "cross-env": "5.2.0",
     "husky": "6.0.0",
     "lint-staged": "11.0.0",
     "mocha": "6.1.4",
+    "sinon": "^11.1.1",
```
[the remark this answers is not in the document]
mattseddon: Probably want to upgrade mocha first v9 has just been released.

src/extension/statusbartoggler.ts
```diff
@@ -43,23 +47,41 @@ export class StatusBarToggler implements Disposable {
         this.update();
     }
 
+    public setCoverage(linePercentage: number | undefined ) {
+        if (Number.isFinite(linePercentage)) {
```
mattseddon, line 51: [F] This let's us differentiate between "No Coverage" (i.e there is no coverage information for this file) and "Coverage 0%" (i.e there is coverage but 0 hits for this file).

src/coverage-system/coverageservice.ts
```diff
@@ -152,24 +150,40 @@ export class CoverageService {
...
 
-    private handleEditorEvents(textEditors: TextEditor[]) {
+    private setStatusBarCoverage(sections: Map<string, Section>, textEditor: TextEditor | undefined ) {
+        try {
+            if (!textEditor) {
+                return this.statusBar.setCoverage(undefined);
+            }
+            const [fileCoverage] = this.sectionFinder.findSectionsForEditor(textEditor, sections);
+            const covered = fileCoverage?.lines?.hit;
+            const total = fileCoverage?.lines?.found;
```
mattseddon, line 165: [Q] Is this the right data / calculation? I was a little unsure of it.

package.json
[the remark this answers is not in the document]
mattseddon: Use of expect from [chai](https://www.chaijs.com/) and the [sinon-chai](https://www.chaijs.com/plugins/sinon-chai/) plugin would be some other good additions to the current testing setup.

src/extension/statusbartoggler.ts
```diff
@@ -52,14 +65,16 @@ export class StatusBarToggler implements Disposable {
...
             this.statusBarItem.command = StatusBarToggler.removeCommand;
-            this.statusBarItem.text = StatusBarToggler.removeText;
+            this.statusBarItem.tooltip = StatusBarToggler.removeWatchToolTip;
+            this.statusBarItem.text = [this.lineCoverage || "No", StatusBarToggler.coverageText].join(" ");
         } else {
             this.statusBarItem.command = StatusBarToggler.watchCommand;
+            this.statusBarItem.tooltip = StatusBarToggler.watchToolTip;
             this.statusBarItem.text = StatusBarToggler.watchText;
         }
         if (this.isLoading) {
```
mattseddon, line 80: [Q] When there is a large coverage file we could end up with `$(loading~spin) No Coverage` for a while. Maybe we should change that text to reflect that coverage is being loaded. WDYT?

mattseddon: We could have the 4 states as:
1. `Watch` for inactive
2. `$(loading~spin) Coverage` while loading 
3. `$(list-ordered) x% Covered` when actively watching and coverage is available for the active editor
4. `$(list-ordered) No Coverage` when actively watching but no coverage is available or there is no active editor 

Lmk what you think.

mattseddon: I went ahead and implemented the above behaviour and separated the code out into a private helper. LMK if you want to see any changes 👍🏻.

Something that I've done / seen elsewhere is the implementing of `$(circle-large-outline)` for "non-loading" state in the UI this fits really well with `$(loading~spin)`. How would you feel about changing `$(list-ordered)` to `$(circle-large-outline)` and having it on all 3 of the non-loading states?

ryanluker: Yah I think these changes make a lot of sense especially the large outline change 👍🏻 .

package.json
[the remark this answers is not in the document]
ryanluker: Gotcha that sounds like a good plan we can get this into the mainline / master and I can iterate on cleaning up those tests 🤔.

src/coverage-system/coverageservice.ts
ryanluker: Yah based on my understanding the spec here https://github.com/mitchhentges/lcov-rs/wiki/File-format#lcov-info-file-output-of-lcov-rs and the discussion information I dup up here https://github.com/ryanluker/vscode-coverage-gutters/discussions/313 I think this is correct hah!

src/extension/statusbartoggler.ts
ryanluker: **Question:** Just for my understanding what do you mean when you add a `[F]` at the front?
I got the others (Q for question, etc) but I was curious about this one haha.

src/coverage-system/coverageservice.ts
```diff
@@ -114,13 +112,13 @@ export class CoverageService {
             `[${Date.now()}][coverageservice]: ${state}`);
     }
 
-    private async loadCacheAndRender() {
+    private async loadCacheAndProcess() {
```
ryanluker, line 115: nice process makes a ton more sense here!

ryanluker (approved):
Great changes here @mattseddon, I took a spin of the PR locally and all seems to be working really well. Only thing I noticed is that I think we are adding the partial coverage to the coverage % in the case of the `example/node` scenario.
image

I am not sure this is bad though and I think it is actually preferrable, maybe just something to be aware of 🤔 . Otherwise this looks ready to merge after you get to looking at my final thoughts!

mattseddon:
@ryanluker I've made [this change](https://github.com/ryanluker/vscode-coverage-gutters/pull/320#discussion_r651438607) to add the circle outline as an idle state, take a look and LMK if you're happy. If not we can roll back and merge / work off the previous commit 👍🏻 .

As for: 
> Great changes here @mattseddon thanks the small video as well. I took a spin of the PR locally and all seems to be working really well. Only thing I noticed is that I think we are adding the partial coverage to the coverage % in the case of the example/node scenario.

I thought this might be the case, that's why I was asking if the calculation was correct. Is there another field in the data that we can use?

mattseddon:
> @ryanluker I've made [this change](https://github.com/ryanluker/vscode-coverage-gutters/pull/320#discussion_r651438607) to add the circle outline as an idle state, take a look and LMK if you're happy. If not we can roll back and merge / work off the previous commit 👍🏻 .
> 
> As for:
> 
> > Great changes here @mattseddon thanks the small video as well. I took a spin of the PR locally and all seems to be working really well. Only thing I noticed is that I think we are adding the partial coverage to the coverage % in the case of the example/node scenario.
> 
> I thought this might be the case, that's why I was asking if the calculation was correct. ~Is there another field in the data that we can use?~ Are the gutters displaying branch coverage? Do we want to match that?

I went back and looked into the code and how the gutters are being calculated, I can see that it is a combination of both the line and branch coverage. I think that it was be a good start to present line coverage (as is) in the status bar and if users asking for something different then we can iterate. WDYT @ryanluker?

ryanluker:
@mattseddon Sounds good we can always iterate once people get their hands on it and see how they like it 🤔 .
Thanks again for the contributions, merging!

Issue #321, opened 2021-06-15
Improve test mocking by adopting sinon


ryanluker:
https://github.com/ryanluker/vscode-coverage-gutters/pull/320#discussion_r650550183
@mattseddon FYI from our discussion in the other PR, I will work on plugging away at fixing and refactoring the existing manual mocks.

### Background
See the past discussion and examples section but this issue will track the refactoring and removal of the manual mocks from the test suite. While useful at the beginning when the codebase and test amount was small, they have not scaled well and it is apparent we need a better tool.

### Prep Work
- [x] Update mocha to v9
- [x] Add sinon-chai https://www.chaijs.com/plugins/sinon-chai/
- [x] Convert `fakeConfig` into an applicable sinon mock
- [x] Remove deprecated assert.equal usage `assert.equal` 
   ```
   /** @deprecated since v9.9.0 - use strictEqual() instead. */
   function equal(actual: any, expected: any, message?: string | Error): void;
   ```

### Refactor existing manual mocks
- [x] `test\coverage-system\coverage.test.ts`
- [x] `test\coverage-system\coverageservice.test.ts`
- [x] `test\extension\config.test.ts`
- [x] `test\files\coverageparser.test.ts`
- [x] `test\files\filesloader.test.ts`

### Past discussions and examples around mocking
https://github.com/ryanluker/vscode-coverage-gutters/pull/270
https://github.com/ryanluker/vscode-coverage-gutters/pull/218
https://github.com/ryanluker/vscode-coverage-gutters/issues/198

PR #322, opened 2021-06-19
Upgrade mocha to v9

mattseddon:
First point from #321.


package.json
```diff
@@ -267,7 +267,7 @@
     "@types/chai": "4.1.7",
     "@types/glob": "7.1.1",
     "@types/lodash": "4.14.126",
-    "@types/mocha": "5.2.6",
+    "@types/mocha": "8.2.2",
```
ryanluker, line 270: I assume no one bothered to update the typed-definitions to 9.x.x 🤔 .
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mocha/index.d.ts

ryanluker (approved):

package.json
mattseddon: Yeah, looks like that's what happened. I went off the latest version available on npm: https://www.npmjs.com/package/@types/mocha

ryanluker:
@mattseddon I also invited you as a collaborator so I don't have to hit the merge button all the time for yah 😉

PR #323, opened 2021-06-26
Start adoption of chai's expect and sinon-chai plugin

mattseddon:
Addresses the second checkbox in #321 and gives a suggestion for addressing the 4th (Remove deprecated assert.equal usage assert.equal).

Adds `sinon-chai`, types and updates `test/extension.test.ts` to `chai`'s `expect` style of assertions (reasoning detailed inline).

This is just a suggestion. LMK what you think.


tsconfig.json
```diff
@@ -4,11 +4,13 @@
         "target": "es6",
         "outDir": "out",
         "alwaysStrict": true,
+		"esModuleInterop": true,
         "lib": [
             "es6",
             "dom",
         ],
         "sourceMap": true,
+		"skipLibCheck": true,
```
mattseddon, line 13: [F] We need to set this options in order to be able to adopt sinon-chai.

tsconfig.json
```diff
@@ -4,11 +4,13 @@
         "target": "es6",
         "outDir": "out",
         "alwaysStrict": true,
+		"esModuleInterop": true,
```
mattseddon, line 7: [F] This just means that we can move away from `import * as` where possible.

test/extension.test.ts
```diff
@@ -1,10 +1,14 @@
-import * as assert from "assert";
+import chai from "chai";
```
mattseddon, line 1: [C] This is the file to focus on. The reason that I'd recommend using the `expect` syntax is that is the closest that we can get to emulating `jest` (most popular Js testing framework) whist using `mocha`.

tsconfig.json
ryanluker: question: what specifically about this causes typescript to dislike sinon-chai?

ryanluker (approved):
Just a few minor questions but looks great to me

tsconfig.json
mattseddon: The error is: 
```
node_modules/@types/sinon-chai/index.d.ts:107:31 - error TS2694: Namespace 'Chai' has no exported member 'ChaiPlugin'.

107 declare const sinonChai: Chai.ChaiPlugin;
```
Much googling reveals that `ChaiPlugin` (for some reason) is private within `Chai`. I will take a look to see if I can find another way around.

mattseddon: very glad you asked about this, I just had to update the types and everything now works.

test/extension.test.ts
```diff
@@ -1,10 +1,14 @@
...
 import { exec } from "child_process";
-import * as sinon from "sinon";
+import sinon from "sinon";
+import sinonChai from "sinon-chai";
 import * as vscode from "vscode";
 import { ICoverageLines } from "../src/coverage-system/renderer";
 import { StatusBarToggler } from "../src/extension/statusbartoggler";
 
+chai.use(sinonChai);
```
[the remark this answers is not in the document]
mattseddon: From [this answer](https://stackoverflow.com/questions/62255953/chai-usechaihttp-once-or-in-every-test-file) it appears that we can have a global fixture but that's not something I've worked with (yet). Maybe a change for the next PR?

mattseddon:
@ryanluker if you're happy with leaving the addition of a test fixture (https://github.com/ryanluker/vscode-coverage-gutters/pull/323#discussion_r659126419) until next weekend then I'll go ahead and merge 👍🏻.

tsconfig.json
ryanluker: Excellent hah!

test/extension.test.ts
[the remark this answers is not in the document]
ryanluker: Nice that makes sense and is much cleaner

ryanluker:
@mattseddon you should be good to merge whenever! I believe with the contributor role you can do that on your own now as well 🤔.

PR #327, opened 2021-06-30
Migrate all test assertions to expect

mattseddon:
Addresses the 4th point in #321.


ryanluker (approved):
Nice looking good to me!

PR #328, opened 2021-07-01
Swap to sinon stub from homerolled fakeConfig

ryanluker:
### Description
Finished part 3 of #321 to use sinon instead of the manual fake. Along with removing the old exportAPI test helper.
cc @mattseddon 

### Design
Originally I was thinking of using sinon mocks but after reading though the documentation [1] it appears that these are more correctly called stubs in sinon land, heh [2].

I also turfed the exportAPI helpers for the extension tests and adopted the sinon spy instead [3], pretty happy with the results!

### Work
- [x] use sinon stub for fakeConfig
- [x] refactor fakeConfig name to stub
- [x] remove exportAPI concept and use sinon spies instead

### Research
1. https://sinonjs.org/releases/v11.1.1/mocks/#when-to-not-use-mocks
2. https://sinonjs.org/releases/v11.1.1/stubs/#when-to-use-stubs
3. https://sinonjs.org/releases/v11.1.1/spies/#using-a-spy-to-wrap-all-object-method
4. https://stackoverflow.com/a/38152098


test/extension.test.ts
```diff
@@ -244,53 +257,53 @@ suite("Extension Tests", function() {
...
         return vscode.commands.executeCommand("coverage-gutters.removeWatch");
     });
 
     test(
         "Run coverage and open files to see line coverage percentage in the status bar @integration",
         async () => {
-        await vscode.commands.executeCommand("workbench.action.closeAllEditors");
-        const setCoverageSpy = sinon.spy(StatusBarToggler.prototype, "setCoverage");
+            await vscode.commands.executeCommand("workbench.action.closeAllEditors");
+            const setCoverageSpy = sinon.spy(StatusBarToggler.prototype, "setCoverage");
```
ryanluker, line 273: @mattseddon Followed along with what you had here but sadly I wasn't able to spy on private typescript functions so I had to change `setDecorationsForEditor` to public 🤔 .

mattseddon: I'll check this out and take a look tonight. 

One quick things that I could suggest is that you can call `restore` in an `afterEach` hook instead of restoring each stub manually.

mattseddon: Actually, you might not be able to do that because of your global stub 🤦🏻‍♂️

ryanluker: @mattseddon Yah I originally had that plan to do the restore / setup in a sandbox inside the before / after.
Wasn't able to figure out how to appease typescript though and it didn't know that the sinon spy was a union type between a `spy | config`.

I ended up just putting them inline as a first pass to get everything working 🤔 .

test/stubs/Config.ts
```diff
@@ -0,0 +1,7 @@
+import sinon from "sinon";
+import { Config } from "../../src/extension/config";
+
+const stubConfig: Config = sinon.stub(Config) as any;
```
mattseddon, line 4: [C] You could use 
```
const stubConfig = sinon.createStubInstance(Config) as Config;
```
here instead.

mattseddon: generally I would create a new one of these for each test because I get worried (and am paranoid) about mutation leading to unexpected results but what you have works and that would be a big change.

test/extension.test.ts
mattseddon: After I switched the stub config to an instance (using `const stubConfig = sinon.createStubInstance(Config) as Config;`) I was able to remove the inline calls to restore and add

```
	this.afterEach(() => restore());
```

at the top of this file and all of the tests still passed.

mattseddon: > @mattseddon Followed along with what you had here but sadly I wasn't able to spy on private typescript functions so I had to change `setDecorationsForEditor` to public 🤔 .

I've had a look into this and you could probably get the visible editors from `window.visibleTextEditors` and then add spies to their `setDecorations` methods. That would give you a true integration test and make it possible to leave the method private. Seeing as there should only be one editor open you could probably even use `window.activeTextEditor` which would be easier.

If you go down that route I would definitely add an `afterEach` hook that runs:

```
commands.executeCommand('workbench.action.closeAllEditors')
```

again this should keep your tests independent. Up to you whether the effort is worth the benefit.

mattseddon (approved):
LGTM boss. Please LMK if you need any clarification 👍🏻.

test/stubs/Config.ts
ryanluker: yah that makes sense, I can easily move this into each test now as a dependency to remove the global nature.

test/extension.test.ts
ryanluker: RE Per test file stubConfig:
That make sense to me, I will try to refine this a bit more, good suggestions!

RE `Public` vs `Private`:
Excellent, I like the idea of using the vscode api for the spies as those have public functions that we use and as you mentioned create better integration tests anyways.

ryanluker: @mattseddon https://github.com/ryanluker/vscode-coverage-gutters/pull/328/commits/e60053cbe07ae8650dbe53532713ec13d0d11b35
RE Config stub:
Ended up cleaning up the config stub a bit more but didn't go the whole way to having it per test case basis.

RE The spy stuff:
Sadly `setDecorations` was not very spyable via sinon on the different vscode provided areas.
I tried a bunch of different options using active and visible editors but nothing would let me spy the decorations setting function, ended up leaving as is for now.

ryanluker:
@mattseddon thanks for the feedback!

PR #329, opened 2021-07-04
Refactor existing manual mocks

### LINKED ISSUE #321: Improve test mocking by adopting sinon

mattseddon:
### Description

Closes #321 by addressing the remaining checkboxes.

I have annotated the PR where I thought that it was appropriate.

As usual LMK if you'd like to see any changes or need clarification on why decisions where made 👍🏻.

### DIFF
--- a/test/coverage-system/coverage.test.ts
+++ b/test/coverage-system/coverage.test.ts
@@ -1,5 +1,6 @@
-import {expect} from "chai";
+import { expect } from "chai";
 import fs from "fs";
+import { afterEach } from "mocha";
 import sinon from "sinon";
 import * as vscode from "vscode";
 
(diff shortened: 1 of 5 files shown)


test/coverage-system/coverage.test.ts
```diff
@@ -8,114 +8,77 @@ import {Config} from "../../src/extension/config";
...
-            .catch(function(error) {
-                if (error.name === "AssertionError") { return done(error); }
-                if (error.message === "could not read from fs") { return done(); }
-                return done(error);
-            });
+        try {
+            await coverage.load("pathtofile");
+        } catch (e) {
+            expect(e.message).to.equal("could not read from fs");
+        }
```
mattseddon, line 33: [F] I have moved all of the tests onto async / await as I find it easier to work with. Seems like all TS is headed in this direction now. LMK what you think.

test/files/coverageparser.test.ts
```diff
@@ -11,41 +20,29 @@ suite("CoverageParser Tests", function() {
         testFiles.set("/file/222", "datastring222");
         testFiles.set("/file/123", "samestringhere");
 
-        // Mock lcovExtract
-        const lcovExtract = async (filename) => {
+        const coverageParsers = new CoverageParser(fakeOutputChannel);
+        sinon.stub(coverageParsers as any, "lcovExtract").callsFake(async (filename) => {
```
mattseddon, line 24: [F] There were two test that I couldn't find away around casting the object to any so that we could stub a private function. This one and `"findCoverageFiles returns an error if no coverage file @unit"`. I think that it is reasonable to leave these as is for now.

test/extension/config.test.ts
```diff
@@ -1,29 +1,32 @@
...
-        },
+    let stubCreateTextEditorDecorationType: sinon.SinonStub;
+    const fakeCreateTextEditorDecorationType = (options: vscode.DecorationRenderOptions) => {
+        expect(Object.keys(options)).to.have.lengthOf(4);
+        return {} as vscode.TextEditorDecorationType;
+    };
 
-        getConfiguration: () => {
+    setup(() => {
+        stubCreateTextEditorDecorationType = sinon.stub(vscode.window, "createTextEditorDecorationType");
```
mattseddon, line 25: [F] This seemed necessary as we want to change the implementation for some tests

test/coverage-system/coverage.test.ts
ryanluker: Nice this vs the promise based approaches seem both equally "good" and match what most developers work with so I think this is fine. As long as we don't start using nested callbacks again 😁 .

test/files/coverageparser.test.ts
ryanluker: Sounds good, we could also use sinon to make the fakes instead of making our own on line 11 but otherwise this makes sense!

ryanluker (approved):
Just a couple questions from myself but otherwise looks great!

test/coverage-system/coverage.test.ts
```diff
@@ -1,4 +1,4 @@
-import {expect} from "chai";
+import { expect } from "chai";
```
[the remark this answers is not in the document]
mattseddon: I think the best course of action would be to switch from tslint to eslint and implement prettier at the same time. That isn't the smallest job though. I'd save it for a later release.

test/files/coverageparser.test.ts
mattseddon: I'll have another look at this but I think that I got a type error when trying to create an output channel stub with sinon.

mattseddon: When trying to create a stub using `sinon.createStubInstance(OutputChannel) as OutputChannel;` the error is `'OutputChannel' only refers to a type, but is being used as a value here.` which will be because we don't have access to the constructor or class. I could stub `vscode.window.createOutputChannel` and have it return our fake object but I don't think it is worth the effort. WDYT @ryanluker?

ryanluker: Yah that is proving tricky, I think this is good enough.

test/coverage-system/coverage.test.ts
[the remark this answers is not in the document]
ryanluker: yah just a thought for future additions, no need to do eslint / prettier in this PR 😁 .

ryanluker:
@mattseddon Looks good to merge to me and then I will do a release next chance I get.

Issue #343, opened 2021-08-21
Switch from sentry to application insights


ryanluker:
### Description
Sentry seems to capture errors between running extensions in the same thread within VS Code.
Swapping to a more standard logging platform for VS Code extensions is preferable to avoid this leakage!

### Work
- [x] Privacy considerations still kept
- [x] Investigate adopting https://github.com/Microsoft/vscode-extension-telemetry
- [x] Implement new error logging capture method

PR #344, opened 2021-08-21
Switch from sentry to application insights

ryanluker:
#343 

### Work
- [x] Refactor crash reporter into generic event reporter
- [x] Deprecated sentry project once released (~30days to let upgrade occur)
- [x] Prevent tests from trigger eventReporter functions


src/extension/crashreporter.ts
```diff
@@ -1,39 +1,21 @@
...
-            });
-            Sentry.configureScope(function(scope) {
-                // Generate a random string for this session
-                // Note: for privacy reason, we cannot fingerprint across sessions
-                scope.setUser({id: uuidv4()});
-            });
-        }
+        this.reporter = new TelemetryReporter(
+            "vscode-coverage-gutters",
+            "2.9.0",
```
mattseddon, line 9: [F] I think that you can get this from `ExtensionContext.extension.packageJSON.version` (In case you don't want to manually enter this each time you bump the version).

mattseddon: I also just realised this is an unsolicited comment, sorry if it is unuseful

ryanluker: @mattseddon I will never say no to feedback! Thanks for the suggestion, looks better to me and saves me having to iterate this all the time 🤔.

ryanluker: @mattseddon I did find that context doesn't have access to this but it looks like it is more standard to use `getExtension()`.
https://github.com/microsoft/vscode-gather/blob/4aca410f888efc0c56e8f45aacda2b8bd0ef3b1b/src/telemetry.ts#L94-L96

ryanluker:
@mattseddon found one issue upstream so we will have to wait on this sadly.
https://github.com/microsoft/vscode-extension-telemetry/pull/69

src/extension/crashreporter.ts
mattseddon: It seems serendipitous that you are doing this now as I am also working on putting analytics in this week. 

I have seen that generally all of the Microsoft repositories copy + paste that same file for dealing with the vscode-extension-telemetry instance (vscode-edge-devtools, vscode-jupyter & vscode-python were the first 3 I checked).

My initial thought was that I was going to go in a different direction and have a single instance of the reporter injected into all of the classes that I needed (hence I thought it would be easy to get the version from `activate(context: ExtensionContext)`). 

After some more thought / digging it appears that I won't actually get that much from doing that so this I'll be going the same route as "everyone" else.

Sorry for the dud suggestion.

ryanluker: @mattseddon interesting, there are some advantages to having one instance though as it is easier to call dispose on later but as you mentioned it seems all of Microsoft's implementations revolve around creating a new reporter each time...

mattseddon: From what I've seen it's actually the same "global" instance for each call of `getTelemetryReporter`. they'll only ever instantiate one unless they clear it:

```
let telemetryReporter: TelemetryReporter | undefined;
function getTelemetryReporter() {
    if (!isTestExecution() && telemetryReporter) {
        return telemetryReporter;
    }
    const extensionId = JVSC_EXTENSION_ID;
    // eslint-disable-next-line @typescript-eslint/no-require-imports
    const extensions = (require('vscode') as typeof import('vscode')).extensions;
    const extension = extensions.getExtension(extensionId)!;
    const extensionVersion = extension.packageJSON.version;

    // eslint-disable-next-line @typescript-eslint/no-require-imports
    const reporter = require('vscode-extension-telemetry').default as typeof TelemetryReporter;
    return (telemetryReporter = new reporter(extensionId, extensionVersion, AppinsightsKey, true));
}
```

We use `@hediet/std/disposable` (because it gives us "hot reload" for dev) which means that we just have to "track" that global reporter somewhere and it will get cleaned up when the parent class gets disposed.

something like:
```
this.dispose.track(getTelementryReporter())
```
in the main `Extension` class.

ryanluker: @mattseddon ahhh gotcha, great explanation! I see what you mean now, by the singleton they got going on there 🤔, crafty! Definitely seems like the way to go for dispose as well given it ultimately just cleans up the config listeners and sends any pending events.
https://github.com/microsoft/vscode-extension-telemetry/blob/main/src/common/baseTelemetryReporter.ts#L319-L325

ryanluker:
@mattseddon Note that there was one last interesting learning
https://github.com/ryanluker/vscode-coverage-gutters/pull/344/commits/1eff40328b16748c52a49996ae5094a932383ec3

Gotta make sure your tests don't trigger events to app insights hah...

ryanluker:
@mattseddon oops! There was actually a couple issues with the above commit I mentioned. Better version below.
https://github.com/ryanluker/vscode-coverage-gutters/commit/9e48acc88bb96e7d416c5e6529f507fc92b08f3a

mattseddon:
> @mattseddon oops! There was actually a couple issues with the above commit I mentioned. Better version below.
> [9e48acc](https://github.com/ryanluker/vscode-coverage-gutters/commit/9e48acc88bb96e7d416c5e6529f507fc92b08f3a)

Thanks, will take a look later on today. Did you consider excluding debug sessions as well?

ryanluker:
@mattseddon Hmmm I wasn't 100% sure how to access that and after looking through the telemetry module they already handled telemetry for yah (both IDE and cli). How where you thinking of excluding the debug sessions?
https://code.visualstudio.com/api/references/vscode-api#env

They don't appear to expose that to us directly like telemetry 🤔 . Maybe we could use the same cli telemetry disable but also here in the `launch.json` https://github.com/ryanluker/vscode-coverage-gutters/blob/master/.vscode/launch.json#L15.

mattseddon:
> They don't appear to expose if something is connected to a debugger unlike telemetry 🤔. Maybe we could use the same cli telemetry disable but also here in the `launch.json` https://github.com/ryanluker/vscode-coverage-gutters/blob/master/.vscode/launch.json#L15.

That should work, I have at least 6 different entry points to launch a VS Code instance that I don't want to send telemetry from so went with environment variables. This the top of our send telemetry function:

```
export const sendTelemetryEvent = <
  P extends IEventNamePropertyMapping,
  E extends keyof P
>(
  eventName: E,
  properties: P[E]
) => {
  if (isTestExecution() || isDebugSession()) {
    return
  }
```

For me this is more explicit (important when you have multiple people working in the code base) but I have to admit what you have looks (is) cleaner.

PR #353, opened 2022-01-05
Could not find coverage file

timthelion:


timthelion:
I'm not a node developer. Is it easy/reasonable to create a temp dir for testing? Currently, this isn't done in the test suit https://github.com/ryanluker/vscode-coverage-gutters/blob/1fa499cf309b7771628bdf78d5e7970ee906ba96/test/files/filesloader.test.ts#L28 but should be both for the happy path and this unhappy one. Can someone help me with that? I don't even know how to interact with the fs in node ;)

timthelion:
Weirdly, this doesn't actually fix it for me when I install the vsix file. But the new error message was enough for me to figure out which dir I had to add to the ignore list. Adding the bad directory to the ignore list did fix this and I think the PR is still totally valid.

ryanluker:
@timthelion Thanks for the PR! I will hopefully have some time this weekend to review it for yeah and hopefully merge it into the next release. Definitely looks like an improvement though on the error message from a first glance 🤔.

src/files/filesloader.ts
```diff
@@ -92,9 +92,16 @@ export class FilesLoader {
                     dot: true,
                     ignore: this.configStore.ignoredPathGlobs,
                     realpath: true,
+                    strict: false,
```
ryanluker, line 95: Just for my understanding, the glob find will continue on error to try to find a possible match correct?
Before this (due to the bug you mentioned) it would stop abruptly if an error occurred even if everything had not yet been searched?

ryanluker (approved):
Looks good! Thanks for the contribution, just one question around the strict flag.

src/files/filesloader.ts
timthelion: Yep, that's right. With strict true it dies if it hits an error.

ryanluker: Excellent, merging this should go out in the 2.9.0 release whenever I get around to finish a couple of the other items.

Issue #364, opened 2022-04-27
Cannot read properties of undefined (reading 'split')


cceckman:
With this [lcov.info.gz](https://github.com/ryanluker/vscode-coverage-gutters/files/8568959/lcov.info.gz) file (uncompressed) in my repository, `coverage-gutters` provides the following output in the VS Code "output" pane- some initial output elided:

```
[1651035766179][coverageservice]: LOADING
[1651035767056][coverageservice]: Loading 1 file(s)
[1651035767056][coverageservice]: /home/cceckman/r/github.com/cceckman/peekatoo/coverage/gcs-70ed0f715011c5c7/lcov.info
[1651035767066][coverageservice]: Loaded 1 data file(s)
[1651035767342][coverageservice]: Caching 1200 coverage(s)
[1651035767342][coverageservice]: READY
[1651035767342][coverageservice]: RENDERING
[1651035767343][watchCoverageAndVisibleEditors]: Cannot read properties of undefined (reading 'split')
[1651035767343][watchCoverageAndVisibleEditors]: TypeError: Cannot read properties of undefined (reading 'split')
    at SectionFinder.resolveFileName (/home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/sectionfinder.js:73:36)
    at SectionFinder.checkSection (/home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/sectionfinder.js:53:38)
    at /home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/sectionfinder.js:24:70
    at Array.filter (<anonymous>)
    at SectionFinder.findSectionsForEditor (/home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/sectionfinder.js:24:45)
    at /home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/renderer.js:32:54
    at Array.forEach (<anonymous>)
    at Renderer.renderCoverage (/home/cceckman/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.9.1/out/src/coverage-system/renderer.js:25:21)
[...]

PR #365, opened 2022-04-27
Address #364: catch errors per-section

cceckman:
In #364, I accidentally produced a file that (apparently) has empty
section filenames, as understood by the extension. This lead to an
undefined property access (`split`) when trying to work with the file
name; that exception propagated all the way up the rendering stack,
resulting in _no_ coverage being displayed.

With this change, if we encounter an error in `checkSection`, we just return
`false` - if it's an erroneous section, it doesn't apply to the current
file. (This is especially true if there is no `section.file` property!)

This way, we still show coverage wherever we can.


src/coverage-system/sectionfinder.ts
```diff
@@ -60,15 +60,22 @@ export class SectionFinder {
...
+            // IE: /var/www/ -> /home/me/
+            const sectionFileName = this.resolveFileName(section.file);
+            if (!isPathAbsolute(sectionFileName)) {
+                return this.checkSectionRelative(sectionFileName, editorFileRelative);
+            } else {
+                return this.checkSectionAbsolute(sectionFileName, editorFileRelative, workspaceFolderName);
+            }
+        } catch (error) {
+            const checkSectionError = `[${Date.now()}][renderer]: ignoring section: error from check: ${error}`;
+            this.outputChannel.appendLine(checkSectionError);
```
ryanluker, line 77: nice job finding the logger and logging this!

src/coverage-system/sectionfinder.ts
```diff
@@ -60,15 +60,22 @@ export class SectionFinder {
      * @param section data section to check against filename
      * @param editorFileRelative normalized relative path (against workspace folder) of editor filename, starts with ###
      * @param workspaceFolderName workspace folder name
+     * @returns true iff this section matches (applies to) the provided editorRelativeFile
```
ryanluker, line 63: super minor: `true iff` -> `true if`

src/coverage-system/sectionfinder.ts
```diff
@@ -60,15 +60,22 @@ export class SectionFinder {
...
      */
     private checkSection(section: Section, editorFileRelative: string, workspaceFolderName: string): boolean {
-        // Check if we need to swap any fragments of the file path with a remote fragment
-        // IE: /var/www/ -> /home/me/
-        const sectionFileName = this.resolveFileName(section.file);
-        if (!isPathAbsolute(sectionFileName)) {
-            return this.checkSectionRelative(sectionFileName, editorFileRelative);
-        } else {
-            return this.checkSectionAbsolute(sectionFileName, editorFileRelative, workspaceFolderName);
+        try {
```
ryanluker, line 66: I think this try catch makes sense here, the only other option is down lower in the check functions or higher in the callers of the `checkSection`.

ryanluker (approved):
One minor thing and then ready to merge!

src/coverage-system/sectionfinder.ts
cceckman: Done.

cceckman:
Thanks for the comments - glad I wasn't too far off base!

ryanluker (approved):

PR #368, opened 2022-05-03
Issue 367 / Switch out tslint for eslint

### LINKED ISSUE #367: Switch out tslint (deprecated) for eslint
Should be an easy swap to eslint from tslint (deprecated).

### Work
- [ ] Install eslint as a dev dependency
- [ ] Use eslint in place of tslint within the project

### Extra info
image
https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration

timagixe:
Closes #367


src/extension/gutters.ts
```diff
@@ -61,6 +61,7 @@ export class Gutters {
                 commands.executeCommand("livePreview.start.internalPreview.atFile", pickedReport, false);
             }
 
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ryanluker, line 64: interesting, given the rule `"@typescript-eslint/no-explicit-any": "off",` I would have thought tslint would ignore these `error: any` types 🤔.

ryanluker:
@timagixe Looking great so far, thanks for tackling this!
I have approved the PR to run against the CI system and also provided a few bits of feedback above.

(we might also need to update the npm scripts to include a few different eslint options based on the CI output🤔)
https://github.com/ryanluker/vscode-coverage-gutters/runs/6274012413?check_suite_focus=true
https://eslint.org/docs/user-guide/command-line-interface#--ext

src/extension/gutters.ts
timagixe:
```json
    "overrides": [
        {
            "files": "**/test/**/*.test.ts",
            "rules": {
                "@typescript-eslint/no-explicit-any": "off",
                "@typescript-eslint/no-empty-function": "off",
                "@typescript-eslint/no-unused-vars": "off"
            }
        }
    ]
```

These three rules above, are turned off only for `test` folder, while they are still applied across the project, i.e. in `src`. That's why I have to manually disable `@typescript-eslint/no-explicit-any` in `gutters.ts` and `coverageparser.ts`.

Let me explain, why I did so, it's because of most of the warnings related to these three rules were produced by `*.test.ts` files.

Please let me know, if this is not desired or we should turn off them across the project instead of only in `test` directory. I will make a proper changes.

timagixe:
@ryanluker thanks for running CI, I have adjusted scripts in https://github.com/ryanluker/vscode-coverage-gutters/pull/368/commits/d00386bfd9f9f187581e9e8da3e818431b518efa. I hope this time all the check will fly 🚀

ryanluker (approved):
Great work! Thanks again for tackling this.

Issue #388, opened 2022-12-12
Change activation event to `workspaceContainers`


SethFalco:
**Describe the bug**

Currently, the extension is set to activate on startup.

![image](https://user-images.githubusercontent.com/22801583/207041148-a198e185-6d59-495d-a54e-abe632e32ca1.png)

I'm looking through the activation events of all my extensions since lately, I've noticed some of them activate eagerly even when they don't apply to my workspace, which makes the startup slower and uses more resources on my system. This adds up when you have many extensions, i.e. I have 92.

VSC provides an activation event for when a certain file pattern is matched in the workspace:

https://code.visualstudio.com/api/references/activation-events#workspaceContains

Do you think it would be feasible to move the activation event for this extension from `*` to `workspaceContains`?

For example:

```json
{
  "activationEvents": [
    "workspaceContains:**/*.xml",
    "workspaceContains:**/lcov.info"
  ]
}
```
> Activate if the workspace contains an XML file, or `lcov.info`.

**Desktop (please complete the following information):**
 - OS: linux
 - Extension Version v2.10.2
 - VSCode Version:

```
Version: 1.75.0-insider
Commit: f076d553210e9ea65e7bccbfa2f7a786c10fcab2
Date: 2022-12-09T09:09:03.185Z
Electron: 19.1.8
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Linux x64 6.0.0-5-amd64 snap
Sandboxed: Yes
```

**Additional context**

* https://github.com/microsoft/vscode/issues/157499

Note, if `workspaceContains` might not be appropriate, we could still consider moving it from `*` to `onStartupFinished`.

> This activation event is emitted and interested extensions will be activated some time after VS Code starts up. This is similar to the `*` activation event, but it will not slow down VS Code startup. Currently, this event is emitted after all the `*` activated extensions have finished activating.
>
> — https://code.visualstudio.com/api/references/activation-events#onStartupFinished

For example:

```json
{
  "activationEvents": [
    "onStartupFinished"
  ]
}
```

ryanluker:
@SethFalco Thanks for the ticket!
Definitely would be better to turf the on startup instantiation of this extension (this was the only option when I initially developed it, I think...).

I like both ideas, but I think it is reasonable to only load the extension if an applicable file is found. I do wonder about how we might consider other file types not defined in the default list.

[...]

PR #393, opened 2023-02-14
chore: change activation event to onstartupfinished

SethFalco:
Changes the activation event from startup to onStartupFinished.

It wasn't having a big impact on startup time (at only around 23 ms) but since it can be put off, figured why not. (I'm trying pretty hard to improve my VSC performance. ^-^')

Read the related issue for more info!

### Related
* Closes https://github.com/ryanluker/vscode-coverage-gutters/issues/388

### DIFF
--- a/package.json
+++ b/package.json
@@ -242,7 +242,7 @@
     }
   },
   "activationEvents": [
-    "*"
+    "onStartupFinished"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


ryanluker (approved):
Looks like the CI system is happy, so I think we are good to go on this!
Thanks again for the contributions in the ticket and PR 🦾 👍🏻 .

PR #406, opened 2023-05-07
Refactored filterLineCoverage() and filterBranchCoverage() in renderer.ts

MylesJP:
### Description
Cleanup some to-dos around using more modern typescript array helpers.

### Work
- [x] Used `.filter()` to reduce number of if-statements and consolidated if-statements where possible.
- [x] Used `.some()` rather than `.find()` to determine if match exists, since we only need boolean return. 
- [x] Potentially faster since `.some()` returns after first match.

### Documentation
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some


ryanluker:
> Interesting that the Windows build of this failed. It seems to pass all tests when I run 'npm run test' on my local machine after installing all dependencies so I think the logic with my change is correct. Any thoughts as to why that is?

Sometimes pulling the vscode image from microsoft's website can fail, and it causes 1/3 build pipelines to implode.
Usually I just hit the retry button as the CI minutes are free for open source projects (afaik hah).

src/coverage-system/renderer.ts
```diff
@@ -116,22 +116,19 @@ export class Renderer {
         if (!section || !section.lines) {
             return;
         }
-        // TODO cleanup this area by using maps, filters, etc
```
ryanluker, line 119: Thanks for tackling these to-dos!

src/coverage-system/renderer.ts
```diff
@@ -143,17 +140,13 @@ export class Renderer {
         if (!section || !section.branches) {
             return;
         }
-        // TODO cleanup this area by using maps, filters, etc
-        section.branches.details.forEach((detail) => {
-            if (detail.taken === 0) {
-                if (detail.line < 0) { return; }
+        section.branches.details.filter((detail) => detail.taken === 0 && detail.line >= 0).forEach((detail) => {
```
ryanluker, line 143: I like this filter, but we might want to put it on multiple lines to aid in readability (if possible).
```ts
  section.branches.details
  .filter((detail) => detail.taken === 0 && detail.line >= 0)
  .forEach((detail) => {
      const partialRange = new Range(detail.line - 1, 0, detail.line - 1, 0);
      // Evaluates to true if at least one element in range is equal to partialRange
      if (coverageLines.full.some((range) => range.isEqual(partialRange))){
          coverageLines.full = coverageLines.full.filter((range) => !range.isEqual(partialRange));
          coverageLines.partial.push(partialRange);
      }
  });
```

src/coverage-system/renderer.ts
```diff
@@ -116,22 +116,19 @@ export class Renderer {
...
-                    coverageLines.none = coverageLines.none.filter((range) => !range.isEqual(lineRange));
-                }
-                coverageLines.full.push(lineRange);
-            } else {
-                const fullExists = coverageLines.full.find((range) => range.isEqual(lineRange));
-                if (!fullExists) {
-                    // only add a none coverage if no full ones exist
-                    coverageLines.none.push(lineRange);
-                }
+        section.lines.details.filter((detail) => detail.line >= 0).forEach((detail) => {
```
ryanluker, line 119: Similar here as below (around re-ability and a small tweak to the indent)
```ts
  section.lines.details
  .filter((detail) => detail.line >= 0)
  .forEach((detail) => {
      const lineRange = new Range(detail.line - 1, 0, detail.line - 1, 0);
      if (detail.hit > 0) {
          // Evaluates to true if at least one element in range is equal to LineRange
          if (coverageLines.none.some((range) => range.isEqual(lineRange))) {
              coverageLines.none = coverageLines.none.filter((range) => !range.isEqual(lineRange))
          }
          coverageLines.full.push(lineRange);
      } else {
          if (!coverageLines.full.some((range) => range.isEqual(lineRange))) {
              // only add a none coverage if no full ones exist
              coverageLines.none.push(lineRange);
          }
      }
  });
```

MylesJP: Makes sense, thanks for the feedback Ryan. I'll make those changes right away.

ryanluker (commented):
Left some initial thoughts and cleaned up the PR body for yeah!

src/coverage-system/renderer.ts
ryanluker: Sweet, thanks! I will try to take a look within the week.

Issue #424, opened 2023-09-11
`Illegal argument: line must be non-negative` when any `detail.line` is 0


sfc-gh-wzhao:
**Describe the bug**
When parsing a coverage dat file where leads to `section.lines.details` or `section.branches.details`'s element in `renderer` will lead to `Illegal argument: line must be non-negative`.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a coverage lcov file containing information at line 0.
2. Enable coverage display

**Expected behaviour**
They should be ignored and not affect display coverage info.

**Screenshots**
Traceback:
```
Error: Illegal argument: line must be non-negative
	at $ (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:10:1101)
	at new n (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:103:4331)
	at new o (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:103:5847)
	at /Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:92:34
	at Array.forEach (<anonymous>)
	at Renderer.filterBranchCoverage (/Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:91:14)
	at /Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:59:18
	at Array.forEach (<anonymous>)
	at Renderer.filterCoverage (/Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:57:18)
	at /Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:36:18
	at Array.forEach (<anonymous>)
	at Renderer.renderCoverage (/Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/renderer.js:25:21)
	at CoverageService.<anonymous> (/Users/****/.vscode/extensions/ryanluker.vscode-coverage-gutters-2.11.0/out/src/coverage-system/coverageservice.js:99:31)
[...]

PR #425, opened 2023-09-11
Fix `Illegal argument: line must be non-negative`

sfc-gh-wzhao:
---
name: Fix `Illegal argument: line must be non-negative`
about: Fix `Illegal argument: line must be non-negative` error by ignoring detail whose line attribute is 0

---

## Issue:
- Fix https://github.com/ryanluker/vscode-coverage-gutters/issues/424

## Changes:
- Change the filter in the renferer to ignore detail whose line attribute is 0.


ryanluker (approved):
Looking good, thanks for the PR!
This will go out with the 2.12.0 release or earlier if we do a smaller minor release sooner.

PR #426, opened 2023-09-17
statusbar reports floor of coverage (resolves #423)

### LINKED ISSUE #423: behavior change suggestion/rfc: statusbar should report `Math.floor` instead of `Math.round`

I open a file with 200 LoC, 1 line uncovered. I expect the summary to _not_ say "100% Coverage," since there's a line missing.

It currently does say 100% though, because $\frac{199}{200} * 100=99.5$, and that rounds to $100$.

If there's interest/folks are okay with it, I think this line:

https://github.com/ryanluker/vscode-coverage-gutters/blob/7b40ddc682dd3d841bf0f14170854ec037992dab/src/coverage-system/coverageservice.ts#L178

should be changed from `Math.round` to `Math.floor`. I'll file a PR for this later today, but wanted to raise the issue too for visibility

* not changing it - but this leads to the current slightly misleading output

**Additional context**

N/A

guppy0130:
---
name: statusbar should report `Math.floor`
about: statusbar should report `Math.floor` instead of `Math.round`

---

## Issue:
- implements proposal in #423

## Changes:
- in a 200 line file with 1 missed line, statusbar would report 100%
  coverage with `Math.round`, which is misleading; there is still one
  missed line, so use `Math.floor` instead (which would report 99%
  coverage)
- @ryanluker

### DIFF
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -175,7 +175,7 @@ export class CoverageService {
             const covered = fileCoverage?.lines?.hit;
             const total = fileCoverage?.lines?.found;
 
-            return this.statusBar.setCoverage(Math.round((covered / total) * 100 ));
+            return this.statusBar.setCoverage(Math.floor((covered / total) * 100 ));
         } catch {
             return this.statusBar.setCoverage(undefined);
         }


ryanluker (approved):
@guppy0130 Thanks for the PR!
Test and code quality workflows pass, so we are good to merge (this will go out in 2.12.0 unless we do a smaller release earlier depending on the other work in that release).

PR #436, opened 2024-01-13
Configure gutter colour the same way as lines

verhovsky:
This is actually possible as we can see from the VS Code source code:

https://github.com/microsoft/vscode/blob/12443ca030f93843ca5e9dfb5682ad92d2c6400c/src/vs/editor/browser/services/abstractCodeEditorService.ts#L801

https://github.com/microsoft/vscode/blob/12443ca030f93843ca5e9dfb5682ad92d2c6400c/src/vs/editor/browser/services/abstractCodeEditorService.ts#L591

This isn't explicitly documented (just the type says "string or Uri") but it is even tested:

https://github.com/microsoft/vscode/blob/12443ca030f93843ca5e9dfb5682ad92d2c6400c/src/vs/editor/test/browser/services/decorationRenderOptions.test.ts#L113


package.json
```diff
@@ -299,8 +299,8 @@
     "@types/sinon-chai": "^3.2.8",
     "@types/uuid": "^8.3.4",
     "@types/vscode": "1.66.0",
-    "@typescript-eslint/eslint-plugin": "^5.22.0",
-    "@typescript-eslint/parser": "^5.22.0",
+    "@typescript-eslint/eslint-plugin": "^6.18.1",
+    "@typescript-eslint/parser": "^6.18.1",
```
verhovsky, line 303: Otherwise you get this error:

```
Parsing error: DeprecationError: 'originalKeywordKind' has been deprecated since v5.0.0 and can no longer be used. Use 'identifierToKeywordKind(identifier)' instead.
```

https://stackoverflow.com/questions/76996326

src/extension/config.ts
```diff
@@ -70,38 +71,36 @@
...
+            if (!colour) {
+                return "";
+            }
+
+            colour = colour
+                .replace('&', '&amp;')
+                .replace("'", '&apos;')
+                .replace('"', '&quot;')
+                .replace('<', '&lt;')
+                .replace('>', '&gt;')
```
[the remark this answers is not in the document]
verhovsky: Indeed

ryanluker:
@verhovsky Thanks for the PR, this will be sweet to have and is much better than the manual icon route.
I do wonder, though, how we will go about releasing this without disrupting the backwards compatibility of the old solution?

I definitely think this should be the new way moving forward, but I wonder if we should leave in the old values (but maybe add deprecated to the description in the `packag.json`, with a mention of them going away in 3.0?

In the meantime, the extension could use an adaptor pattern to allow us to move forward with the better approach and still provide support for the old file method so we can release this in 2.x? (which will be much sooner than a hypothetical 3.0 heh).

src/extension/config.ts
```diff
@@ -70,38 +71,37 @@ export class Config {
...
+                .replace(/"/g, '&quot;')
+                .replace(/</g, '&lt;')
+                .replace(/>/g, '&gt;')
+                .replace(/\r\n/g, '&#13;')
+                .replace(/[\r\n]/g, '&#13;');
+            if (!colour) {
+                return "";
+            }
+
+            const svg = '<svg width="32" height="48" viewPort="0 0 32 48" xmlns="http://www.w3.org/2000/svg"><polygon points="16,0 32,0 32,48 16,48" fill="' + colour + '"/></svg>';
```
ryanluker, line 88: I think the tests will also need to be adjusted to support this (or if we support the old method and the new, then we could add a new test to cover this).
https://github.com/ryanluker/vscode-coverage-gutters/actions/runs/7951747196/job/21742231483?pr=436#step:6:128

verhovsky:
This would only matter to people that only set a gutter color and no line color and people that have a different gutter color than the line highlight color. I can't imagine anyone going through the trouble of generating an SVG file just to change the color of a gutter in your code editor, surely not many people use this setting.

ryanluker:
> This would only matter to people that have a different gutter color than the line highlight color and people that only set a gutter color and no line color. I can't imagine anyone going through the trouble of generating an image file just to change the color of a gutter in their code editor, surely not many people use this setting.

Alright I agree with that logic, makes sense!
We can also have an addition in extension marketplace readme mentioning a migration path.

package.json
ryanluker: gotcha thanks!

src/extension/config.ts
```diff
@@ -70,38 +71,37 @@ export class Config {
...
+
+            const icon = 'data:image/svg+xml;base64,' + Buffer.from(svg).toString('base64');
+            return Uri.parse(icon);
```
ryanluker, line 91: Optional Logging: This is looking good, I do wonder though if we should add some logging to assist with if the make icon process fails in some way 🤔.

```ts
    const outputChannel = vscode.window.createOutputChannel("coverage-gutters");
    const configStore = new Config(context);
```

test/extension/config.test.ts
```diff
@@ -91,13 +92,15 @@ suite("Config Tests", () => {
...
-        iconPathLight = "";
+        highlightdark = "rgb(255, 0, 0)";
+        highlightlight = "rgb(0, 0, 255)";
+
+        const preamble = /^image\/svg\+xml;base64,/;
         stubCreateTextEditorDecorationType.callsFake((options) => {
-            expect((options.dark as any).gutterIconPath).to.equal(iconPathDark);
-            expect((options.light as any).gutterIconPath).to.include("./app_images/");
+            expect(((options.dark as any).gutterIconPath as any).path).to.be.a('string').and.match(preamble).and.satisfy((icn: string) => atob(icn.replace(preamble, '')).includes(highlightdark));
+            expect(((options.light as any).gutterIconPath as any).path).to.be.a('string').and.match(preamble).and.satisfy((icn: string) => atob(icn.replace(preamble, '')).includes(highlightlight));
```
ryanluker, line 103: Props 👍🏻 : Cool test, thanks for adding this!
>   ✔ Should set the gutter icon colour to the provided value if set @unit

ryanluker (approved):
Sweet feature addition @verhovsky , thanks for contributing!

ryanluker:
@verhovsky FYI this is going out with today's release. I was just doing the QA as a double check, and it was working nicely locally for me! Thanks again for the contribution.
image

Issue #228, opened 2019-07-17
Coverage Service Not Loading


dgrahn:
**Describe the bug**
Coverage service is not loading.

**To Reproduce**
1. Open VS Code
2. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>8</kbd>
3. View error log.

```
[1563372232910][coverageservice]: INITIALIZING
[1563372244566][coverageservice]: LOADING
```

**Expected behaviour**
Coverage will load.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Extension Version: 2.4.0
 - VSCode Version: 1.32.1

**Additional context**
Using pytest-cov to generate `cov.xml`

ryanluker:
@dgrahn Thanks for submitting an issue! Can you confirm that, if you checkout the example project and run the coverage https://github.com/ryanluker/vscode-coverage-gutters/tree/master/example/python, you get the same no coverage result?

ryanluker:
Closing this due to age but willing to open again if the issue persists!

dgrahn:
Ok. I apologize that this has languished.

I just downloaded that project. I see `foobar/.coverage` and `foobar/cov.xml`. Whenever I try loading the service, I get this message.

```
[1566388888477][coverageservice]: LOADING
```

dgrahn:
Is there a way to enable verbose logging?

ryanluker:
@dgrahn verbose logging is on my default and you should be seeing more information from the example project (see logs file below). I have also attached a video showing the example project in action (depending on where you opened it you might have issues with finding the coverage).
https://imgur.com/a/WUzzkKL
```
[1566748736670][coverageservice]: INITIALIZING
[1566748755827][coverageservice]: LOADING
[1566748755878][coverageservice]: Loading 5 file(s)
[1566748755881][coverageservice]: Loaded 5 data file(s)
[1566748755915][coverageservice]: Caching 12 coverage(s)
[1566748755915][printDataCoverage]: Coverage -> 12
[1566748755915][coverageservice]: READY
[1566748755915][coverageservice]: RENDERING
[1566748755915][coverageservice]: READY
[1566748759451][coverageservice]: RENDERING
[1566748759452][renderer][section file path]: C:\dev\vscode-coverage-gutters\example\python\python\foobar/tests/bar/a.py
[1566748760440][coverageservice]: READY
```

ryanluker:
[...]

PR #445, opened 2024-07-13
Ignore hidden directories by default within the glob pattern

ryanluker:
### Description
Based on a suggestion here https://github.com/ryanluker/vscode-coverage-gutters/issues/228#issuecomment-2185276726 around ignoring hidden directories / folders by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
         "coverage-gutters.ignoredPathGlobs": {
           "type": "string",
           "scope": "resource",
-          "default": "**/{node_modules,venv,.venv,vendor}/**",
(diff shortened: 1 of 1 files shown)


ryanluker:
Tried the easiest option https://github.com/ryanluker/vscode-coverage-gutters/pull/445/commits/d66d8d7d0eeb408a222fb33a78c366720f10808f but I think we need to go the more advance route of writing custom ignore functions (similar to what I posted here https://github.com/ryanluker/vscode-coverage-gutters/issues/228#issuecomment-2226993189). Definitely requires new test coverage though and not just a one line change sadly 😅.

Issue #212, opened 2019-03-18
Add webpack to the build pipeline


ryanluker:
## Work
- Integration webpack into the extension build flow
- Add applicable typescript bundling bits
- Add performance improvements as directed by the docs below 
- https://code.visualstudio.com/api/working-with-extensions/bundling-extension
- https://code.visualstudio.com/api/working-with-extensions/bundling-extension#migrate-an-existing-extension
- https://docs.sentry.io/platforms/javascript/sourcemaps/#webpack

PR #467, opened 2025-01-29
Add esbuild to bundle and package the extension

ryanluker:
### Work
#212 
Nothing too crazy here, mainly just following along with the migration guide / vscode extension development documentation.
https://code.visualstudio.com/api/working-with-extensions/bundling-extension#migrate-an-existing-extension

#### Tested the following areas
- vsce package (installed and confirmed it can display coverage)
- npm run test
- npm run build

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         include:
           - os: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
       with:
         node-version: 22.x
     - run: npm install
+    - run: npm run build
     - run: xvfb-run -a npm test
       if: runner.os == 'Linux'
     - run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,9 +13,9 @@ jobs:
       matrix:
         include:
           - os: macos-latest
-            test: test
+            test: test-headless
           - os: windows-latest
-            test: test
+            test: test-headless
           - os: ubuntu-latest
             test: test-headless
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
           - os: windows-latest
             test: test
           - os: ubuntu-latest
-            test: test
+            test: test-headless
     runs-on: ${{ matrix.os }}
     env:
       CI: true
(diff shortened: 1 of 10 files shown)


ryanluker:
@mattseddon I came across some interesting issues with the `main` value from the `package.json` and running the test suite. I tried to change the value to `"main": "./dist/extension",` as per the vscode docs.

They have this snippet, where it looks like this should be the main entry for the compiled extentions?
https://code.visualstudio.com/api/working-with-extensions/bundling-extension#run-the-extension
> Before you can run the extension, the main property in package.json must point to the bundle, which for the configuration above is ["./dist/extension"](https://github.com/microsoft/vscode-references-view/blob/d649d01d369e338bbe70c86e03f28269cbf87027/package.json#L26). With that change, the extension can now be executed and tested.

https://code.visualstudio.com/api/references/extension-manifest
> main 		string 	The entry point to your extension.

Should I be able to add something to the mocha `runTests` file that allows me to use a different `main` value for testing vs for running / packaging?

mattseddon:
> @mattseddon I came across some interesting issues with the `main` value from the `package.json` and running the test suite. I tried to change the value to `"main": "./dist/extension",` as per the vscode docs.
> 
> They have this snippet, where it looks like this should be the main entry for the compiled extentions? https://code.visualstudio.com/api/working-with-extensions/bundling-extension#run-the-extension
> 
> > Before you can run the extension, the main property in package.json must point to the bundle, which for the configuration above is ["./dist/extension"](https://github.com/microsoft/vscode-references-view/blob/d649d01d369e338bbe70c86e03f28269cbf87027/package.json#L26). With that change, the extension can now be executed and tested.
> 
> https://code.visualstudio.com/api/references/extension-manifest
> 
> > main 		string 	The entry point to your extension.
> 
> Should I be able to add something to the mocha `runTests` file that allows me to use a different `main` value for testing vs for running / packaging? (Or maybe I am miss understanding what the `main` manifest value is for)

If you're moving the `out` directory to `dist` you'll need to repoint the test command as well: `"test": "npm run compile && node ./out/test/runTest.js",`

mattseddon:
I think my previous comment stands, if you are going to move `main` onto `dist` then all of the scripts need to point to `dist`, probably best to get rid of `tsc` based builds altogether and to use `watch` from `esbuild`.

ryanluker:
> I think my previous comment stands, if you are going to move `main` onto `dist` then all of the scripts need to point to `dist`, probably best to get rid of `tsc` based builds altogether and to use `watch` from `esbuild`.

Gotcha! Thanks for the info.

ryanluker:
Victory over the build ci runner! Let us hope for success in the next marketplace deploy 🙏🏻.

PR #480, opened 2025-06-04
Make ms-vscode.live-server an optional dependency

### LINKED ISSUE #392: Does this extension really need Live Preview?
Hi! I have Live Preview and when trying to disable Live Preview, I get this warning that this extension depends on it. Is this a bug or an actual dependency?

image

Gerrit0:
Removes hard dependency on live-server extension, replacing it with an error message if the preview command is used when not installed.

Resolves #392


test/runTest.ts
```diff
@@ -1,48 +1,57 @@
...
 async function main() {
     try {
+        const vscodeVersion = "insiders";
         const extensionDevelopmentPath = path.resolve(__dirname, "../../");
         const extensionTestsPath = path.resolve(__dirname, "index");
 
-        // Add the dependent extension for test coverage preview functionality
-        const vscodeExecutablePath = await downloadAndUnzipVSCode("insiders");
-        const [cliPath, ...args] = resolveCliArgsFromVSCodeExecutablePath(vscodeExecutablePath);
+        // Ensure live-server is not installed
```
Gerrit0, line 13: I'm not much of a fan of this as it means that we end up running all the tests twice, so this roughly doubles the amount of time it takes to run the tests, but wasn't able to find a better way of running a test without the live server extension installed...

test/runTest.ts
```diff
@@ -1,48 +1,57 @@
...
+            extensionDevelopmentPath,
+            extensionTestsPath,
+            launchArgs: ["example/example.code-workspace"],
+            extensionTestsEnv: {
+                TEST_FILTER: "no-live-server"
+            }
+        })
 
-        // Useful for debugging failing dependant extension installs
-        console.info(output);
```
Gerrit0, line 32: The runVSCodeCommand function will throw if calling VSCode fails, so this will be caught by the catch block now

test/runTest.ts
mattseddon: If you wrap `const livePreview = extensions.getExtension("ms-vscode.live-server");` in a function, you can stub that function in a couple of tests and get rid of all of this.

mattseddon: or leave it installed and only stub it being missing in one test

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -8,13 +8,13 @@ about: Create a report to help us improve
...
 **To Reproduce**
-Steps to reproduce the behaviour:
+Steps to reproduce the behavior:
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
 4. See error
 
-**Expected behaviour**
+**Expected behavior**
```
mattseddon, line 17: Can you move these unrelated changes out of this PR, please.

Gerrit0: Gone

test/createTestPackageJson.ts
```diff
@@ -0,0 +1,15 @@
+import path from "path";
```
Gerrit0, line 1: This is from #482

ryanluker:
@Gerrit0 Thanks for the PR and thanks @mattseddon for reviewing it, looking great so far.

I merged the other PR with the test changes / fixups, so I have rerun the CI action for yeah Gerrit.

test/runTest.ts
ryanluker: Excellent! This is a super nice cleanup 🦾 .

src/extension/gutters.ts
```diff
@@ -29,6 +30,15 @@ export class Gutters {
 
     public async previewCoverageReport() {
         try {
+            const livePreview = this.getLiveServerExtension();
+            if (!livePreview) {
+                await window.showErrorMessage("Live Preview extension not installed", {
+                    modal: true,
+                    detail: "The ms-vscode.live-server extension must be installed to preview the coverage report."
+                }, "Ok");
```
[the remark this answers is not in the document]
Gerrit0: I considered that, but decided not to open that can of worms at this point given that there are other messages which contain hardcoded English text:

- <https://github.com/ryanluker/vscode-coverage-gutters/blob/be757eac82bdfaae5627684d5e84952f28614585/src/files/filesloader.ts#L22>
- <https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/files/filesloader.ts#L108>
- <https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/extension/gutters.ts#L38>
- <https://github.com/ryanluker/vscode-coverage-gutters/blob/master/src/extension/statusbartoggler.ts#L21-L22>

ryanluker (approved):
Excellent improvement to the codebase, extra extensions that you shouldn't need is super annoying, so this is very welcome.

PR #482, opened 2025-06-11
copy (and update) package.json to out directory to enable mocking in integration tests

### LINKED ISSUE #479: Broken extension integration tests
**Describe the bug**

The "Run coverage and open files to see line coverage percentage in the status bar \@integration" test does not actually perform a test, and appears to be broken.

**To Reproduce**
1. Replace the incomplete `expect` calls with complete expect calls in the unit test

```diff
@@ -353,7 +375,7 @@ suite("Extension Tests", function () {
             setCoverageSpy.resetHistory();
             await vscode.window.showTextDocument(testJSDocument, vscode.ViewColumn.One);
 
-            expect(setCoverageSpy.calledWith(84))
+            expect(setCoverageSpy.calledWith(84)).to.equal(true);
```

2. Run tests, they fail.

**Expected behavior**

The extension tests should pass

mattseddon:
Closes #479 

The underlying issue is that sinon cannot mock modules in the bundled `dist` folder. In order for this to work we need to copy the package.json from the root folder into the `out/` folder (generated by `compile-tests`) and update the `main` entry to point at `src/extension`.

I have taken the opportunity to update the tests in `extension.tests.ts` to ensure they are now testing the correct things.

### DIFF
--- a/src/coverage-system/coverageservice.ts
+++ b/src/coverage-system/coverageservice.ts
@@ -194,8 +194,10 @@ export class CoverageService {
     }
 
     private listenToEditorEvents() {
-        this.editorWatcher = window.onDidChangeVisibleTextEditors(
+        this.editorWatcher = window.onDidChangeActiveTextEditor(
             this.handleEditorEvents.bind(this),
         );
+
+
     }
 }
--- a/package.json
+++ b/package.json
@@ -243,7 +243,7 @@
     "pretty": "eslint --fix --ext .ts",
     "pre-commit": "lint-staged",
     "compile-tests": "tsc -p . --outDir out",
-    "pretest": "npm run compile-tests",
+    "pretest": "npm run compile-tests && node ./out/test/createTestPackageJson.js",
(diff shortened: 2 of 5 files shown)


src/coverage-system/coverageservice.ts
```diff
@@ -194,8 +194,10 @@ export class CoverageService {
     }
 
     private listenToEditorEvents() {
-        this.editorWatcher = window.onDidChangeVisibleTextEditors(
+        this.editorWatcher = window.onDidChangeActiveTextEditor(
```
mattseddon, line 197: I found this bug whilst fixing the tests

Gerrit0:
This doesn't appear to work. I tried to stub a method as suggested in https://github.com/ryanluker/vscode-coverage-gutters/pull/480#discussion_r2139106954, which didn't work. After applying your changes here, the stubs are still not correctly applied.

ryanluker (approved):
Looking great to me, thanks for your work on this!

PR #483, opened 2025-06-11
use coverage filename in cache keys

### LINKED ISSUE #481: Not handling multiple lcov files correctly
Due to the amount of tests we have, we have multiple test suites that run separately, each generates its own subfolder under `coverage/`.  Each focuses on certain types of tests or sections of code, but there's some overlap.  

In cases where two lcov files contain coverage information about a line, but one says it's covered and one says it isn't, the plugin is sometimes showing it as uncovered.  If I delete the lcov file that _doesn't_ cover the line, then the plugin will correctly show it as covered.  

Looking at the cache code, I'm assuming the Map keys are overlapping between these lcov files, it's just random which Map key is 'winning'

mattseddon:
Closes #481 

cc @ryanluker - I'm leaving this one as draft as I'm not too sure how you've setup the data structures here.


test/files/coverageparser.test.ts
```diff
@@ -8,28 +8,26 @@ import { CoverageParser } from "../../src/files/coverageparser";
...
 
-    test("filesToSections properly deduplicates coverages @unit", async () => {
-        // Setup a map of test keys and data strings
-        // Note: we include a duplicate key to test deduplication
+    test("filesToSections properly combines coverages @unit", async () => {
         const testFiles = new Map();
-        testFiles.set("/file/123", "datastringhere");
-        testFiles.set("/file/111", "datastring111");
-        testFiles.set("/file/222", "datastring222");
-        testFiles.set("/file/123", "samestringhere");
```
mattseddon, line 20: As this is a map it only ever has 3 entries, making the test meaningless :(

ryanluker: Whoops hah! Darn Ryan from many ages ago 😅.

src/files/coverageparser.ts
```diff
@@ -54,11 +54,12 @@ export class CoverageParser {
     }
 
     private async convertSectionsToMap(
+        coverageFilename: string,
         data: Section[],
     ): Promise<Map<string, Section>> {
         const sections = new Map<string, Section>();
         const addToSectionsMap = async (section: Section) => {
-            sections.set(section.title + "::" + section.file, section);
+            sections.set([coverageFilename, section.title, section.file].join("::"), section);
```
ryanluker, line 62: I am curious how this will decide between two conflicting coverage files / lines? Do we just need to assume the "best case" scenario and pick the coverage line that is more positive?

src/files/coverageparser.ts
```diff
@@ -80,7 +81,7 @@ export class CoverageParser {
             try {
                 parseContentCobertura(xmlFile, async (err, data) => {
                     checkError(err);
-                    const sections = await this.convertSectionsToMap(data);
+                    const sections = await this.convertSectionsToMap(coverageFilename,data);
```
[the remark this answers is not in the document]
mattseddon: If I use format on save, I end up with a whole load of unrelated changes in the file.

example/node/lcov.info
```diff
@@ -1,10058 +0,0 @@
-TN:
```
mattseddon, line 1: @ryanluker er.... why was this file so long?

src/files/coverageparser.ts
```diff
@@ -17,147 +17,293 @@ export class CoverageParser {
...
 
-            // file is an array
-            let coverage = new Map<string, Section>();
-
             // get coverage file type
             const coverageFile = new CoverageFile(fileContent);
             switch (coverageFile.type) {
                 case CoverageType.CLOVER:
-                    coverage = await this.xmlExtractClover(fileName, fileContent);
+                    await this.xmlExtractClover(
```
mattseddon, line 33: [F] Each of these `extract` methods now mutate the original `coverages` Map.

src/files/coverageparser.ts
```diff
@@ -17,147 +17,293 @@ export class CoverageParser {
...
-            sections.set(section.title + "::" + section.file, section);
+            const key = [section.title, section.file].join("::");
+            const existingSection = coverages.get(key);
+
+            if (!existingSection) {
+                coverages.set(key, section);
+                return;
+            }
+
+            const mergedSection = this.mergeSections(existingSection, section);
```
mattseddon, line 76: [F] If we already have a section with the current title and file, then we merge the line and branch information as they are used downstream.

src/files/coverageparser.ts
```diff
@@ -17,147 +17,293 @@ export class CoverageParser {
...
                 });
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
             } catch (error: any) {
                 checkError(error);
             }
         });
     }
 
+    private mergeSections(existingSection: Section, section: Section): Section {
```
mattseddon, line 186: [F] If you look at https://github.com/ryanluker/vscode-coverage-gutters/pull/483/commits/66df5d7bc3a00d76dd0406072a339c23e194cc56 you can see how I initially worked this out by updating the `node` example project.

test/files/coverageparser.test.ts
```diff
@@ -8,36 +8,122 @@ import { CoverageParser } from "../../src/files/coverageparser";
...
+            "./integration/lcov.info",
+            `TN:
+            SF:${fileUnderTest}
+            FN:1,test
+            FNF:1
+            FNH:1
+            FNDA:1,test
+            DA:1,1
+            DA:2,1
+            DA:3,1
```
mattseddon, line 60: Line 3 is only covered by the integration test in this example.

src/files/coverageparser.ts
ryanluker: Gotcha that makes sense to allow for the multiple coverage files scenario.

src/files/coverageparser.ts
```diff
@@ -1,10 +1,10 @@
...
-import {Section, source} from "lcov-parse";
-import {OutputChannel} from "vscode";
+import { parseContent as parseContentClover } from "@cvrg-report/clover-json";
+import { parseContent as parseContentCobertura } from "cobertura-parse";
+import { parseContent as parseContentJacoco } from "@7sean68/jacoco-parse";
+import { Section, source } from "lcov-parse";
+import { OutputChannel } from "vscode";
 
-import {CoverageFile, CoverageType} from "./coveragefile";
+import { CoverageFile, CoverageType } from "./coveragefile";
```
ryanluker, line 7: Yeahhh this file hasn't been touched in a while, didn't even have the latest lintings 😓.

src/files/coverageparser.ts
```diff
@@ -17,147 +17,289 @@ export class CoverageParser {
...
+        const branches = this.mergeBranchCoverage(existingSection, section);
+
+        return {
+            ...existingSection,
+            lines,
+            branches,
+        };
+    }
+
+    private mergeLineCoverage(
```
ryanluker, line 193: This merging logic looks correct to me, nice work 🦾.

PR #491, opened 2025-09-12
Enables a hit count indicator to show number of times a line has been hit

ogechno:
This feature enables users to view the amount of times a line has been hit during test execution. This closes the discussion in issue #194

# Design Rationale
In issue #194, there was the discussion around having a GitLens-style hit counter at the end of each line. I decided against that approach since usually you want to quickly skim the coverage report and spot any lines with low hit counts during execution (e.g. when reviewing fuzzing coverage reports). This would be difficult if the hit counter is at the end of the line.
That's why I went with displaying the hit counts to the left of the code, similar to how lcov and grcov are doing it. This makes it much easier to scan through the report and immediately see which lines might need more test coverage, rather than having to read across each line to find the counts at the end

# Screenshot
Screenshot 2025-09-12 at 16 33 41


ryanluker (commented):
Looking great so far! We will need a few tests as well and maybe a way to "cap" the hit count if it is greater than x (not sure if the count matters if you are at >1000 hits, hah).

src/coverage-system/renderer.ts
```diff
@@ -139,7 +139,8 @@ export class Renderer {
             }
             
             const range = new Range(lineNumber, 0, lineNumber, 0);
-            const paddedHitCount = hitCount.toString().padStart(paddingWidth, '\u00A0');
+            const displayValue = hitCount > 1000 ? '>1000' : hitCount.toString();
+            const paddedHitCount = displayValue.padStart(paddingWidth, '\u00A0');
```
ryanluker, line 143: Another option for really high hit counts, would be to use something like `k` | `m` to denote thousands / millions. (This was suggested by someone else here: https://github.com/ryanluker/vscode-coverage-gutters/issues/505#issuecomment-4216657912)

mattseddon: If you have a test suite that hits a line close to a million times... you might be doing it wrong 😂

ogechno: @ryanluker Great idea.

@mattseddon There are definitely cases where a line would be hit more than a million times, e.g., [Google fuzzing OpenSSL](https://storage.googleapis.com/oss-fuzz-coverage/openssl/reports/20260805/linux/src/openssl/ssl/ssl_ciph.c.html#L121 ).

test/runTest.ts
```diff
@@ -18,6 +20,7 @@ async function main() {
         console.info("Success!");
         process.exit(0);
     } catch (err) {
+        if (!retryEnabled) return process.exit(1);
```
ryanluker: @ogechno Added this env flag so we can run the tests locally without needing to wait for the 3-4 retry loops.

test/extension/config.test.ts
```diff
@@ -99,6 +115,10 @@ suite("Config Tests", () => {
 
         const preamble = /^image\/svg\+xml;base64,/;
         stubCreateTextEditorDecorationType.callsFake((options) => {
+            // Ignore hitCount decorations
+            if (options.before) {
+                return {} as vscode.TextEditorDecorationType;
+            }
```
ryanluker: @ogechno Most of the test failures came from the fact that we have two types of decorations now and the old mocks / spys didn't know about this new data possibility when considering decorations.

`gutter hits`:
```ts
            before: {
                color: hitCountColor,
            }
```

`coverage stylings`:
```ts
            dark: {
                backgroundColor: showLineCoverage ? noCoverageDarkBackgroundColour : "",
                gutterIconPath: showGutterCoverage ? makeIcon(noCoverageDarkBackgroundColour) : "",
                overviewRulerColor: showRulerCoverage ? noCoverageDarkBackgroundColour : "",
            },
            isWholeLine: true,
            light: {
                backgroundColor: showLineCoverage ? noCoverageLightBackgroundColour : "",
                gutterIconPath: showGutterCoverage ? makeIcon(noCoverageLightBackgroundColour) : "",
                overviewRulerColor: showRulerCoverage ? noCoverageLightBackgroundColour : "",
            },
            overviewRulerLane: OverviewRulerLane.Full,
```

ryanluker:
@ogechno I got all the tests fixed up but didn't have time today to add any new ones for your new feature.

Issue #345, opened 2021-09-01
Gcov support


Jaakkonen:
**Is your feature request related to a problem? Please describe.**
Currently it's not possible to see coverage of programs written in C.

**Describe the solution you'd like**
Add support for formats used by `gcov` and llvm profiling.

**Describe alternatives you've considered**

https://github.com/JacquesLucke/gcov-viewer but that doesn't have same kind of auto detection and less-obstructing UI that coverage-gutters has.
One option is to use CLI/TUI programs to do this but it's really nice to be able to edit code and see coverage at the same time.

**Additional context**
Guide for getting coverage with LLVM: https://releases.llvm.org/12.0.1/tools/clang/docs/SourceBasedCodeCoverage.html (this is not gcov and should have its own feature request ticket)
Gcov trivial example: https://github.com/shenxianpeng/gcov-example
Ok-ish explanation how gcov works: https://www.tutorialspoint.com/unix_commands/gcov.htm
https://llvm.org/docs/CommandGuide/llvm-profdata.html

ryanluker:
@Jaakkonen Thanks for the issue! I am unfamiliar with the coverage landscape for C programs but gcov looks to be one popular standard 🤔 but we do not currently support it (as you alluded to above).

I believe people have used the expension with gcov by using lcov to convert the coverage?
```
So here we will use lcov tool to check line coverage and generate html reports . For this command is as shown below.
lcov --coverage --directory . --output-file main_coverage.info
This will create a coverage report by taking .gcno and .gcda files from ‘.’ i.e. current folder where these files are present and generate an output file with coverage info in file called main_coverage.info .
Now we have coverage info lets display it in html file using below command.
```
https://medium.com/@naveen.maltesh/generating-code-coverage-report-using-gnu-gcov-lcov-ee54a4de3f11

Let me know if that helps any in the meantime but we could consider gcov support in the future (or if there is already a javscript / typescript package that can convert gcov -> lcov then that would make this addition even easier)!

Ali-Amir:
You can generate `xml` with gcvor today and that worked for me without having to do format conversion.

Note that you may have issues with relative paths depending on how your coverage was generated. You can fix that with `remotePathResolve` though. For example, in my coverage xml I'm getting `<source>..</source>` and had to do setup `remotePathResolve`:
```
"coverage-gutters.remotePathResolve": [
    "../src",
    "./src",
],
```

PR #498, opened 2025-12-21
Feature Enhancement: per-file coverage in Explorer, C/C++ Support & visualize branch/region execution

### LINKED ISSUE #151: Is it possible to display coverage percentage next to files in the file explorer?
When opening a coverage report, e.g. from `nyc`, I first get an overview of coverage for files and folders. It would be nice to see this percentage next to the files in the file explorer (if that is even possible).

### LINKED ISSUE #391: Display information about partial branch coverage
Some lines are yellow (partial), but I don't know which branch was taken and which one still needs testing.

I would like to have a tooltip that displays additional coverage information.

Currently, this only is available with other tools.

mr-u0b0dy:
## Features
- Explorer coverage badges: per-file coverage percentages with color-coded thresholds (≥80% green, ≥50% yellow, <50% red); configurable via `showExplorerCoverage`.
- C/C++ coverage support: parse LLVM-cov JSON and visualize branch/region execution through CodeLens and Hover providers.
- Hover enhancements: highlight the active coverage region and show the closest region’s execution count; include Cobertura per-line condition coverage details.

## Issues
- Closes #497 
- Closes #391
- Closes #151
- Partially addresses #345

image


@types/lcov-parse/index.d.ts
```diff
@@ -10,7 +10,9 @@ declare namespace parse {
         block: number,
         branch: number,
         line: number,
-        taken: number
+        taken: number,
+        condition_coverage?: number,  // percentage of condition branches taken (0-100)
+        missing_branches?: number[]   // line numbers of untaken branches
```
ryanluker, line 15: @mr-u0b0dy Should this be a part of the branch detail or its own new `RegionDetail` interface?

I am getting up to speed on the documentation, so please correct me if I am making incorrect assumptions 🙏🏻.

https://llvm.org/docs/CoverageMappingFormat.html#mapping-region

example/cpp/README.md
```diff
@@ -0,0 +1,66 @@
+# C++ Coverage Example (LLVM-cov JSON)
```
ryanluker, line 1: @mr-u0b0dy Excellent work on these two new example projects for c / c++! I will try them out locally to confirm the instructions are able to generate some region coverage for us.

example/rust/README.md
```diff
@@ -0,0 +1,71 @@
+# Rust Coverage Example
```
ryanluker, line 1: Excellent, a Rust example was on my holiday todo list!

src/files/coveragefile.ts
```diff
@@ -4,6 +4,7 @@ export enum CoverageType {
     CLOVER,
     COBERTURA,
     JACOCO,
+    LLVM_COV_JSON, // LLVM coverage JSON format
```
ryanluker: We probably don't need the comment here (I would also shorten the enum to just LLVM to match the others).


```suggestion
    LLVM,
```

src/files/coveragefile.ts
```diff
@@ -31,6 +32,17 @@ export class CoverageFile {
             possibleType = CoverageType.JACOCO;
         } else if (file.includes("<?xml")) {
             possibleType = CoverageType.COBERTURA;
+        } else if (file.trim().startsWith("{") || file.trim().startsWith("[")) {
```
ryanluker, line 35: Probably fine, but I was hoping there could be something more specific for us to look at in the file 🤔.
https://llvm.org/docs/CoverageMappingFormat.html#llvm-ir-representation

Based on the documentation above, this might be the best we can depend on.

**----EDIT----**
One area that we could maybe tinker with, is this "metadata" section near the end that includes versioning details?
```py
"type":"llvm.coverage.json.export","version":"2.0.1"}
```

mr-u0b0dy:
I’ve made separate commits for each feature addition, and I hope this makes the review easier. Since these are incremental commits—where upcoming changes depend on files introduced in the first commit—it might be better to include them in a single PR. I’ll proceed with fixing the test failures, and I’m also looking forward to the addition of the line hit count feature (#491).

ryanluker:
> I’ve made separate commits for each feature addition, and I hope this makes the review easier. Since these are incremental commits—where upcoming changes depend on files introduced in the first commit—it might be better to include them in a single PR. I’ll proceed with fixing the test failures, and I’m also looking forward to the addition of the line hit count feature (#491).

Alright no worries on the mega PR this time, there might be some process improvements we can do in the future around leveraging stacked PRs for big code additions like this 🤔.
https://abhinav.github.io/git-spice/

mr-u0b0dy:
Makes sense — stacked PRs sound like a good fit for changes like this. I’ll take a look at git-spice and keep that in mind for future work. Thanks!

ryanluker:
> Makes sense — stacked PRs sound like a good fit for changes like this. I’ll take a look at git-spice and keep that in mind for future work. Thanks!

Review progress wise, I should be able to finish up the review this upcoming weekend.

mr-u0b0dy:
@ryanluker Any update on this PR?

ryanluker:
> @ryanluker Any update on this PR?

Thanks for the ping, I haven't got much further due to the size of the PR and truthfully I am not that interested in reviewing / adopting too much AI generated code into the extension. That isn't something you did wrong though as there isn't anything in the [contributing](https://github.com/ryanluker/vscode-coverage-gutters/blob/master/CONTRIBUTING.md) guidelines around AI use.

I had a partial review done that I submitted and I can try to come back in the future and look more for yeah.

example/cpp/README.md
mr-u0b0dy: Thanks! Let me know if the instructions don't reproduce the region coverage on your machine and I'll adjust.

src/files/coveragefile.ts
mr-u0b0dy: Applied, thanks. Heads up: the suggestion commit renamed the enum member but left three CoverageType.LLVM_COV_JSON references behind, so npm run check-types was failing on master. Fixed in 6c021fd.

src/files/coverageparser.ts
```diff
@@ -94,14 +103,167 @@ export class CoverageParser {
...
+                let index = 0;
+                while (index < xmlFile.length) {
+                    // Find next class or line
+                    classOpenRegex.lastIndex = index;
+                    const classMatch = classOpenRegex.exec(xmlFile);
+                    if (classMatch && (classMatch.index >= index)) {
+                        currentFilename = classMatch[1];
+                        if (!coberturaConditionsByFile.has(currentFilename)) {
+                            coberturaConditionsByFile.set(currentFilename, {});
+                        }
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd. The walker now looks ahead for both the next `<class>` and the next `<line>` and takes whichever comes first, so each class keeps its own lines instead of only the last one in the report.

Writing the regression test for this turned up a second bug in the same scanner. The `<line>` pattern required a closing `</line>`, and its optional `condition-coverage` group sat behind a lazy `[^>]*?` that was free to match empty — so the group never fired and self-closing lines swallowed the content of later ones. Against `example/c/coverage.xml` it matched 29 of 88 lines and captured condition coverage **zero** times, meaning the Cobertura condition hover had always been showing `0% (0/0)`.

Attributes are now read independently of their order and self-closing `<line ... />` is handled. The integration test asserts the real values (`50%`, `3/6`) rather than just `> 0`.

src/coverage-system/branchcoverageproviders.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * Highlight a specific region range in the editor
+     */
+    public highlightRegion(editor: vscode.TextEditor, startLine: number, startCol: number, endLine: number, endCol: number) {
+        const range = new vscode.Range(
+            new vscode.Position(startLine - 1, startCol - 1),
+            new vscode.Position(endLine - 1, endCol - 1)
+        );
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd. Columns are clamped with `Math.max(0, value - 1)` before the `Position` is built, so a column 0 segment lands on character 0 instead of throwing. Added a unit test that highlights a region at column 0.

src/coverage-system/coverageservice.ts
```diff
@@ -60,6 +68,40 @@ export class CoverageService {
...
+    }
+
+    public clearProvidersData() {
+        if (this.branchCoverageCodeLensProvider) {
+            this.branchCoverageCodeLensProvider.clearCoverageData();
+        }
+        if (this.branchCoverageHoverProvider) {
+            this.branchCoverageHoverProvider.clearCoverageData();
+        }
+    }
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd. `clearProvidersData()` now hands the decoration provider an empty cache, which fires its change event and drops the explorer badges along with the gutters.

src/files/coveragefile.ts
```diff
@@ -31,6 +32,17 @@ export class CoverageFile {
...
+            // LLVM coverage JSON format (starts with { or [)
+            try {
+                JSON.parse(file);
+                possibleType = CoverageType.LLVM_COV_JSON;
+            } catch {
+                // Not valid JSON, treat as LCOV
+                if (file !== "") {
+                    possibleType = CoverageType.LCOV;
+                }
+            }
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd. Detection is schema-based now rather than "is it JSON": the `llvm.coverage.json.export` type tag first, then the `data[].files[]` shape for older exports that predate the tag, then the gcovr `files[].file` shape. Anything else falls through to lcov as before.

Added tests for a gcovr JSON report (detected) and an istanbul-shaped document (must not be claimed).

src/coverage-system/branchcoverageproviders.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+}
+
+/**
+ * Provides hover information for branch coverage details with region highlighting
+ */
+export class BranchCoverageHoverProvider implements vscode.HoverProvider {
+    private coverageData: Map<string, Section> = new Map();
+    private regionHighlighter: RegionHighlighter;
+    private lastHoverPosition: { line: number; col: number } | undefined;
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd — removed.

src/coverage-system/coverage.ts
```diff
@@ -30,23 +30,24 @@ export class Coverage {
...
-                {placeHolder},
+                { placeHolder },
             );
-            if (!item) {
-                window.showWarningMessage("Did not choose a file!");
-                return;
-            }
+            const timeoutPromise = new Promise<QuickPickItem | undefined>((resolve) => {
+                setTimeout(() => resolve(undefined), autoPickTimeoutMs);
+            });
```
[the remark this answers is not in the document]
mr-u0b0dy: Fixed in 6c021fd. The timeout handle is kept and cleared in a `finally` once the race settles, so a quick pick never leaves a pending timer behind.

src/coverage-system/branchcoverageproviders.ts
```diff
@@ -0,0 +1,336 @@
...
+            ).length;
+            const percentage = Math.round((takenBranches / totalBranches) * 100);
+
+            const codeLens = new vscode.CodeLens(
+                range,
+                {
+                    title: `${takenBranches}/${totalBranches} branches taken (${percentage}%)`,
+                    tooltip: `Branch coverage: ${takenBranches} out of ${totalBranches} branches executed`,
+                    command: "",
+                }
```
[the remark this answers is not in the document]
mr-u0b0dy: Not taking this one — I think it is a false positive.

An empty command id is the supported way to render an informational, non-clickable lens. VS Code's `codelensWidget.ts` emits an `<a>` only when `lens.command.id` is truthy and otherwise falls back to a plain `<span>`, and in that branch the lens is never registered in the widget's `_commands` map — so a `command '' not found` error is not reachable from here.

Omitting `command` would be the worse option: `CodeLens.isResolved` becomes false, VS Code then calls `resolveCodeLens`, and this provider does not implement it, so the lens would sit unresolved.

@types/lcov-parse/index.d.ts
mr-u0b0dy: Good question, and it's worth separating two things that got conflated here. `condition_coverage` and `missing_branches` are Cobertura condition data (from `<line condition-coverage="50% (3/6)">`), not LLVM mapping regions — so `RegionDetail` would be a misnomer for them. The actual LLVM region data doesn't go through `BranchDetail` at all; it's attached to the section as `__llvmSegmentsByLine`. Both metadata blobs are currently untyped properties cast onto Section rather than declared interfaces. Happy to extract a proper typed module for them if you'd like.

src/files/coveragefile.ts
mr-u0b0dy: Your edit note was the right instinct. Detection now checks the schema instead of "is it JSON": the `llvm.coverage.json.export` type tag first, then the `data[].files[]` shape for older exports that predate the tag, then the gcovr `files[].file` shape. Anything else falls through to lcov. Tests added for gcovr JSON.

PR #503, opened 2026-01-30
Enhance reading & loading & rendering performance

JackKuo-tw:
To be honest, the code changes are all finished by Claude Sonnet 4.5 and Gemini 3 Pro.
I made this change because this plugin run slow on very large project, and I have to share computing resource with many colleagues.

I did test it locally without issue. If you think it's improper to use AI generated code, feel free to reject this PR 😊

### DIFF
--- a/src/coverage-system/renderer.ts
+++ b/src/coverage-system/renderer.ts
@@ -39,12 +39,11 @@ export class Renderer {
         sections: Map<string, Section>,
         textEditors: readonly TextEditor[],
     ) {
+        // Single-pass iteration for better performance
         textEditors.forEach((textEditor) => {
             // Remove all decorations first to prevent graphical issues
             this.removeDecorationsForEditor(textEditor);
-        });
 
-        textEditors.forEach((textEditor) => {
             // Reset lines for new editor
             const coverageSets: CoverageSets = {
                 full: new Set<number>(),
--- a/src/coverage-system/renderer.ts
+++ b/src/coverage-system/renderer.ts
@@ -12,6 +12,12 @@ export interface ICoverageLines {
     none: Range[];
 }
 
+interface CoverageSets {
+    full: Set<number>;
+    partial: Set<number>;
+    none: Set<number>;
+}
+
 export class Renderer {
     private configStore: Config;
     private sectionFinder: SectionFinder;
@@ -33,12 +39,6 @@ export class Renderer {
         sections: Map<string, Section>,
         textEditors: readonly TextEditor[],
     ) {
-        const coverageLines: ICoverageLines = {
-            full: [],
-            none: [],
-            partial: [],
-        };
-
         textEditors.forEach((textEditor) => {
             // Remove all decorations first to prevent graphical issues
             this.removeDecorationsForEditor(textEditor);
@@ -46,16 +46,25 @@ export class Renderer {
 
         textEditors.forEach((textEditor) => {
             // Reset lines for new editor
-            coverageLines.full = [];
-            coverageLines.none = [];
-            coverageLines.partial = [];
+            const coverageSets: CoverageSets = {
+                full: new Set<number>(),
+                none: new Set<number>(),
+                partial: new Set<number>(),
+            };
 
             // find the section(s) (or undefined) by looking relatively at each workspace
             // users can also optional use absolute instead of relative for this
             const foundSections = this.sectionFinder.findSectionsForEditor(textEditor, sections);
             if (!foundSections.length) { return; }
 
-            this.filterCoverage(foundSections, coverageLines);
+            this.filterCoverage(foundSections, coverageSets);
+
+            const coverageLines: ICoverageLines = {
+                full: this.setsToRanges(coverageSets.full),
+                none: this.setsToRanges(coverageSets.none),
+                partial: this.setsToRanges(coverageSets.partial),
+            };
+
             this.setDecorationsForEditor(textEditor, coverageLines);
         });
     }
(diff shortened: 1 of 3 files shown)


ryanluker:
> To be honest, the code changes are all finished by Claude Sonnet 4.5 and Gemini 3 Pro. I made this change because this plugin run slow on very large project, and I have to share computing resource with many colleagues.
> 
> I did test it locally without issue. If you think it's improper to use AI generated code, feel free to reject this PR 😊

Thanks for the ticket and the honesty on using AI, much appreciated.

> If you think it's improper to use AI generated code, feel free to reject this PR 😊

We don't have a policy written out anywhere on the repo (other than the [CONTIBUTING.md](https://github.com/ryanluker/vscode-coverage-gutters/blob/master/CONTRIBUTING.md) file) but as long as we can understanding and adapt the code to match the patterns, styles and systems in the repository I don't see any issues!

> I made this change because this plugin run slow on very large project

Thanks, I will take a look at the PR today!

src/coverage-system/renderer.ts
```diff
@@ -33,29 +39,31 @@ export class Renderer {
...
 
-        textEditors.forEach((textEditor) => {
             // Reset lines for new editor
-            coverageLines.full = [];
-            coverageLines.none = [];
-            coverageLines.partial = [];
+            const coverageSets: CoverageSets = {
+                full: new Set<number>(),
+                none: new Set<number>(),
+                partial: new Set<number>(),
```
ryanluker, line 51: I guess it makes sense for these to be sets as you can only have one unique coverage type per line (there are scenarios where you can have multiple coverage types on the same line though).

I don't believe we have this ability yet (via the extension) but it might come in a future PR / ticket.
https://github.com/ryanluker/vscode-coverage-gutters/issues/497
https://llvm.org/docs/CoverageMappingFormat.html#mapping-region

src/coverage-system/renderer.ts
```diff
@@ -33,29 +39,31 @@ export class Renderer {
         sections: Map<string, Section>,
         textEditors: readonly TextEditor[],
     ) {
-        const coverageLines: ICoverageLines = {
-            full: [],
-            none: [],
-            partial: [],
-        };
-
+        // Single-pass iteration for better performance
```
ryanluker, line 42: I haven't done any performance tuning for the extension in a long while, I will have to refresh my memory on some of the debugging logs I have used in the past (what method did you use to judge the performance improvement?).

src/coverage-system/renderer.ts
```diff
@@ -97,60 +105,66 @@ export class Renderer {
...
-                // Evaluates to true if at least one element in range is equal to LineRange
-                if (coverageLines.none.some((range) => range.isEqual(lineRange))) {
-                    coverageLines.none = coverageLines.none.filter((range) => !range.isEqual(lineRange))
-                }
-                coverageLines.full.push(lineRange);
-            } else {
-                if (!coverageLines.full.some((range) => range.isEqual(lineRange))) {
-                    // only add a none coverage if no full ones exist
-                    coverageLines.none.push(lineRange);
+            .filter((detail) => detail.line > 0)
```
ryanluker, line 128: Minor code styling: We probably want to leave this as is and use the code formatter we have set up in the pre-commit hooks. (If you are already using that then great, we can disregard my comment).

src/coverage-system/renderer.ts
```diff
@@ -97,60 +105,66 @@ export class Renderer {
...
+            .forEach((detail) => {
+                const line = detail.line - 1;
+                if (detail.hit > 0) {
+                    if (coverageSets.none.has(line)) {
+                        coverageSets.none.delete(line);
+                    }
+                    coverageSets.full.add(line);
```
ryanluker, line 135: This cleaner section replaces the logic below, which looks to be better / equivalent in my eyes?
```ts
                // Evaluates to true if at least one element in range is equal to LineRange
                if (coverageLines.none.some((range) => range.isEqual(lineRange))) {
                    coverageLines.none = coverageLines.none.filter((range) => !range.isEqual(lineRange))
                }
                coverageLines.full.push(lineRange);
```

src/coverage-system/renderer.ts
```diff
@@ -97,60 +105,66 @@ export class Renderer {
...
-                coverageLines.full.push(lineRange);
-            } else {
-                if (!coverageLines.full.some((range) => range.isEqual(lineRange))) {
-                    // only add a none coverage if no full ones exist
-                    coverageLines.none.push(lineRange);
+            .filter((detail) => detail.line > 0)
+            .forEach((detail) => {
+                const line = detail.line - 1;
+                if (detail.hit > 0) {
+                    if (coverageSets.none.has(line)) {
```
ryanluker, line 132: We might want to add back the comment (or update it slightly) so we remember that this helps remove the none coverage if there is atleast some type of coverage hit.

src/coverage-system/renderer.ts
```diff
@@ -97,60 +105,66 @@ export class Renderer {
...
+                        coverageSets.none.delete(line);
+                    }
+                    coverageSets.full.add(line);
+                } else {
+                    if (!coverageSets.full.has(line)) {
+                        // only add a none coverage if no full ones exist
+                        coverageSets.none.add(line);
+                    }
```
ryanluker, line 140: Nice rework @JackKuo-tw (+Claude 🤖)

src/coverage-system/renderer.ts
```diff
@@ -97,60 +105,66 @@ export class Renderer {
...
-        .forEach((detail) => {
-            const partialRange = new Range(detail.line - 1, 0, detail.line - 1, 0);
-            // Evaluates to true if at least one element in range is equal to partialRange
-            if (coverageLines.full.some((range) => range.isEqual(partialRange))){
-                coverageLines.full = coverageLines.full.filter((range) => !range.isEqual(partialRange));
-                coverageLines.partial.push(partialRange);
-            }
+            .filter((detail) => detail.taken === 0 && detail.line > 0)
+            .forEach((detail) => {
+                const line = detail.line - 1;
```
ryanluker, line 155: We probably want to add back the comment and update the variable to be about partial coverage again.

src/files/coverageparser.ts
```diff
@@ -21,39 +21,42 @@ export class CoverageParser {
...
             const coverageFile = new CoverageFile(fileContent);
             switch (coverageFile.type) {
                 case CoverageType.CLOVER:
-                    await this.xmlExtractClover(
-                        coverages,
+                    parsePromises.push(this.xmlExtractClover(
                         fileName,
                         fileContent
-                    );
+                    ));
```
ryanluker, line 34: Personal preference, but if the code linter / reformatter doesn't care then we can ignore this.
```suggestion
                    parsePromises.push(
                        this.xmlExtractClover(
                            fileName,
                            fileContent
                        )
                    );
```

src/files/coverageparser.ts
```diff
@@ -21,39 +21,42 @@ export class CoverageParser {
...
                 case CoverageType.LCOV:
-                    this.lcovExtract(coverages, fileName, fileContent);
+                    parsePromises.push(this.lcovExtract(fileName, fileContent));
                     break;
                 default:
                     break;
             }
         }
+
+        const results = await Promise.all(parsePromises);
```
ryanluker, line 56: Ok cool, this will have all the promises race each other instead of doing the awaits in series like we had before.

src/files/coverageparser.ts
```diff
@@ -21,39 +21,42 @@ export class CoverageParser {
...
+        const flattenedSections = results.reduce((acc, val) => acc.concat(val), []);
+        await this.addSections(coverages, flattenedSections);
```
ryanluker, line 58: Ok this makes sense, it moves the coverage adding to the very end instead of per file 👍🏻.

src/files/coverageparser.ts
```diff
@@ -21,39 +21,42 @@ export class CoverageParser {
...
-                    this.lcovExtract(coverages, fileName, fileContent);
+                    parsePromises.push(this.lcovExtract(fileName, fileContent));
                     break;
                 default:
                     break;
             }
         }
+
+        const results = await Promise.all(parsePromises);
+        const flattenedSections = results.reduce((acc, val) => acc.concat(val), []);
```
ryanluker, line 57: Not 100% sure what this is doing, maybe a small comment here would help?

src/files/filesloader.ts
```diff
@@ -41,28 +45,27 @@ export class FilesLoader {
      * @param files files that are to turned into data strings
      */
     public async loadDataFiles(files: Set<string>): Promise<Map<string, string>> {
-        // Load the files and convert into data strings
-        const dataFiles = new Map<string, string>();
-        for (const file of files) {
-            dataFiles.set(file, await this.load(file));
-        }
-        return dataFiles;
+        // Load all files in parallel for better performance
```
ryanluker, line 48: Loading files like this makes sense to me 👍🏻 .

src/files/filesloader.ts
```diff
@@ -41,28 +45,27 @@ export class FilesLoader {
...
-    private load(path: string) {
-        return new Promise<string>((resolve, reject) => {
-            readFile(path, (err, data) => {
-                if (err) { return reject(err); }
-                return resolve(data.toString());
-            });
-        });
+    private async load(path: string): Promise<string> {
+        const data = await readFile(path, 'utf-8');
+        return data;
```
ryanluker, line 60: Did we lose the error handling from the return of `readFile`?

src/files/filesloader.ts
```diff
@@ -88,32 +91,27 @@ export class FilesLoader {
...
-                            window.showWarningMessage(`An error occured while looking for the coverage file ${err}`);
-                        }
-                        return resolve(new Set());
-                    }
-                    const setFiles = new Set<string>();
-                    files.forEach((file) => setFiles.add(file));
-                    return resolve(setFiles);
-                },
-            );
-        });
```
ryanluker, line 117: Seems fine to promisify this glob usage, but I took a quick look at the docs, and it seems like the library already supports promises / async calling? https://github.com/isaacs/node-glob?tab=readme-ov-file#globpattern-string--string-options-globoptions--promisestring--path

Maybe we need to update the package to get this functionality 🤔.
https://github.com/ryanluker/vscode-coverage-gutters/blob/master/package.json#L299

ryanluker (commented):
@JackKuo-tw Thanks again for the PR, I gave some initial insights and approved your PR for building via the CI runner.

PR #508, opened 2026-03-09
Add retry option to launchArgs in runTest

ryanluker:
Attempt to add automatic retries to the ci builds to prevent failure email spam.

### Extra info:
https://mochajs.org/declaring/retrying-tests/


ryanluker:
Trying the retry a different way.

PR #524, opened 2026-07-25
Reject invalid UTF-8 coverage files

### LINKED ISSUE #504: Loading `.lcov` files with an encoding different than UTF-8 silently fails.
**Describe the bug**

Coverage loading from non-UTF-8 `.lcov` files silently fails and shows no coverage.
It might seem like non-UTF-8 `.lcov` files are a bit unlikely to occur in nature, but when doing `llvm-cov export ... > coverage.lcov` on Windows, the output file will not have UTF-8 encoding.

A workaround is to open the file and save it as UTF-8.

**To Reproduce**
Steps to reproduce the behaviour:
1. Save your `.lcov` file that's read by the extension with an encoding that's not UTF-8.
2. Try to visualize the coverage with this extension. Look at extension logs.

**Expected behaviour**
An error is signalled to the user that the coverage file could not be read.

**Screenshots**
No screenshots, but here are the logs when reading a file that's not UTF-8.
```
[1771233780941][coverageservice]: Loaded 1 data file(s)
[1771233780941][coverageservice]: Caching 0 coverage(s)
[1771233780941][coverageservice]: READY
[1771233780942][coverageservice]: RENDERING
[1771233780942][coverageservice]: READY
```

**Desktop (please complete the following information):**
 - OS: windows
 - Extension Version 2.14.0
 - VSCode Version 1.109.3

**Additional context**
N/A

yangfan-yf-yf:
## Issue

Fixes #504.

## Changes

Coverage files are now decoded as strict UTF-8. Previously, invalid bytes were replaced during decoding, which could make the LCOV parser produce an empty coverage result without reporting the encoding problem. Invalid input now stops loading and returns a path-specific error through the existing error path.

`ignoreBOM: true` preserves the previous behavior for UTF-8 files that start with a byte-order mark.

## Validation

- Added a regression test with invalid UTF-8 bytes.
- `MOCHA_GREP='FilesLoader Tests' npm test` — 5 passing.
- `npm run check-types`
- `npm run lint`
- `npm run compile`
- `git diff --check`


src/files/filesloader.ts
```diff
@@ -53,7 +54,14 @@ export class FilesLoader {
         return new Promise<string>((resolve, reject) => {
             readFile(path, (err, data) => {
                 if (err) { return reject(err); }
-                return resolve(data.toString());
+                try {
+                    return resolve(new TextDecoder(
+                        "utf-8",
+                        {fatal: true, ignoreBOM: true},
+                    ).decode(data));
```
ryanluker, line 61: This makes sense to me, just one question below.

Did you choose to include `ignoreBOM` as an extra nice addition?
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#ignorebom

**EDIT**
Never-mind, I saw you mentioned this in the PR body, makes sense!
> ignoreBOM: true preserves the previous behavior for UTF-8 files that start with a byte-order mark.

test/files/filesloader.test.ts
```diff
@@ -28,6 +28,27 @@ suite("FilesLoader Tests", () => {
 
     });
 
+    test("loadDataFiles rejects coverage data that is not valid UTF-8 @unit", async () => {
```
ryanluker, line 31: Thanks for adding a test, much appreciated!

ryanluker (approved):
Just a couple questions but looking pretty ready to merge, thanks for the contribution!

ryanluker:
@yangfan-yf-yf The failing macos build is fixed in another branch I am working on.
https://github.com/ryanluker/vscode-coverage-gutters/pull/526

